PR #1, opened 2023-01-06
ValidatingAdmissionPolicy for C-0038

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/host-pid-ipc-privileges/raw.rego


slashben:
Merged it, there is a small fix to be done in the police name

suhasgumma:
> Merged it, there is a small fix to be done in the police name

Forgot to change that. Will update in the next PR.

PR #3, opened 2023-01-09
ValidatingAdmissionPolicy for C-0044

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/container-hostPort/raw.rego

### DIFF
--- a/controls/C-0001/tests.json
+++ b/controls/C-0001/tests.json
@@ -8,7 +8,7 @@
         ]
     },
     {
-        "name": "Pod without image from quay.io is blocked",
+        "name": "Pod without image from quay.io is allowed",
         "template": "pod.yaml",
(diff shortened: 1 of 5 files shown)


suhasgumma:
Hi @slashben, I have added ports to `pod.yaml` and `deployment.yaml` files in the test directory because I could not create lists using the current script for tests. Please let me know if there is any way or if we have to modify the script.

Thank you.

slashben:
> Hi @slashben, I have added ports to `pod.yaml` and `deployment.yaml` files in the test directory because I could not create lists using the current script for tests. Please let me know if there is any way or if we have to modify the script.
> 
> Thank you.

It is ok, I will check why adding lists did not work!

slashben (approved):

PR #6, opened 2023-01-09
ValidatingAdmissionPolicy for C-0017

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/immutable-container-filesystem/raw.rego

### DIFF
--- a/controls/C-0017/policy.yaml
+++ b/controls/C-0017/policy.yaml
@@ -1,0 +1,27 @@
+apiVersion: admissionregistration.k8s.io/v1alpha1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "kubescape-c-0017-deny-mutable-container-filesystem"
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   [""]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["pods"]
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments","replicasets","daemonsets","statefulsets"]
+    - apiGroups:   ["batch"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["jobs","cronjobs"]
+  validations:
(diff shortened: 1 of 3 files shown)


controls/C-0017/policy.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments","replicasets","daemonsets","statefulsets"]
+    - apiGroups:   ["batch"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["jobs","cronjobs"]
+  validations:
+    - expression: "object.kind != 'Pod' || object.spec.containers.all(container, container.securityContext.readOnlyRootFilesystem == true)"
```
slashben, line 22: This might not work every time for the following reason.

This field does not required in the object. If you do not specify it, the default behavior is `readOnlyRootFilesystem = false`. 

Neither the `securityContext` nor the `readOnlyRootFilesystem` are required to be defined. In this case, since the default behavior is what the policy blocks, this works to some extent since if the Pod does not have `securityContext` or `readOnlyRootFilesystem`, the CEL evaluation fails and the Pod is denied. 

The expression should be something like this:
```
object.kind != 'Pod' || object.spec.containers.all(container, has(container,securityContext) &&  has(container.securityContext,readOnlyRootFilesystem) && container.securityContext.readOnlyRootFilesystem == true)
```
In this case we explicitly check that these fields exist in the object. 

Again, from the functional perspective your version works. But there is an implicit assumption that might not work in other controls.

slashben: Sorry, the `has` macro works like this (not as above) `has(container.securityContext)`

suhasgumma: > Neither the `securityContext` nor the `readOnlyRootFilesystem` are required to be defined. In this case, since the default behavior is what the policy blocks, this works to some extent since if the Pod does not have `securityContext` or `readOnlyRootFilesystem`, the CEL evaluation fails and the Pod is denied.

I have ignored checking existence for this reason. I understand that it is not a best practice. I will update the expression to check for existence first.

slashben (approved):

PR #7, opened 2023-01-10
ValidatingAdmissionPolicy for C-0005

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/insecure-port-flag/raw.rego


slashben:
Hey @suhasgumma , this control is not useful for the admission policy. This control is checking the API server configuration which is the very same entity that enforces these policies. Somewhat like a  🐔 and 🥚 situation 😄

PR #8, opened 2023-01-10
ValidatingAdmissionPolicy for C-0004

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/resources-memory-limit-and-request/raw.rego


suhasgumma:
Hi @slashben! Is there any particular reason to set the type of `memoryRequestMin, memoryRequestMax` to `array` in `CRD`? If not, shall I change them to integer type?

slashben (approved):

slashben:
You are right, I tried to understand why this was defined as an array of integers. You can changed this to a scalar

PR #11, opened 2023-01-11
ValidatingAdmissionPolicy for C-0013

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/non-root-containers/raw.rego

### DIFF
--- a/controls/C-0013/policy.yaml
+++ b/controls/C-0013/policy.yaml
@@ -20,43 +20,175 @@ spec:
       resources:   ["jobs","cronjobs"]
   validations:
     - expression: > 
-        object.kind != 'Pod' || (object.spec.containers.all(container, has(container.securityContext) && has(container.securityContext.allowPrivilegeEscalation) && container.securityContext.allowPrivilegeEscalation == false)) && (
-          (has(object.spec.securityContext) && (
-            (has(object.spec.securityContext.runAsNonRoot) && object.spec.securityContext.runAsNonRoot == true) ||
-            (has(object.spec.securityContext.runAsUser) && object.spec.securityContext.runAsUser != 0)
-          )) || (
-            object.spec.containers.all(container, has(container.securityContext) && (
-              (has(container.securityContext.runAsNonRoot) && container.securityContext.runAsNonRoot == true) ||
-              (has(container.securityContext.runAsUser) && container.securityContext.runAsUser != 0)
-            ))
+        object.kind != 'Pod' || object.spec.containers.all(container,
+          (
+            (
+              has(container.securityContext) && 
+              has(container.securityContext.allowPrivilegeEscalation) && 
+              container.securityContext.allowPrivilegeEscalation == false
+            ) ||
+            (
+              (
+                !has(container.securityContext) || !has(container.securityContext.allowPrivilegeEscalation)
+              ) &&
+              (
+                has(object.spec.securityContext) && 
+                has(object.spec.securityContext.allowPrivilegeEscalation) && 
+                object.spec.securityContext.allowPrivilegeEscalation == false
+              )
+            )
+          ) &&
+          (
+            (
+              (
+                has(container.securityContext) && 
+                has(container.securityContext.runAsNonRoot) && 
+                container.securityContext.runAsNonRoot == true
+              ) ||
+              (
+                (
+                  !has(container.securityContext) || !has(container.securityContext.runAsNonRoot)
+                ) &&
+                (
+                  has(object.spec.securityContext) && 
+                  has(object.spec.securityContext.runAsNonRoot) && 
+                  object.spec.securityContext.runAsNonRoot == true
+                )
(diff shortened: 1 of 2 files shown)


suhasgumma:
There is one more issue. `PodSecurityContext` does not support `allowPrivilegeEscalation`. 
For more details, see this issue: https://github.com/kubescape/regolibrary/issues/186

suhasgumma:
Keeping that in mind, I've written CEL such that it directly checks if each container set the `securityContext.allowPrivilegeEscalation` to `false`. Even if one container did not set it, the control returns `false`.

slashben:
This is a very complex control, I will review a bit later

Daniel-GrunbergerCA:
Hi @suhasgumma 
Please note that the `if` and `else` statements are evaluated in order (https://www.openpolicyagent.org/docs/latest/policy-language/#else-keyword).
That means that first we check the container security context for the given fields, and only if they are not present there, we check the Pod security context. The reason for that being that the container security context takes precedence over the Pod security context (https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#securitycontext-v1-core).
This behavior needs to be reflected on your check as well. For example, it is possible that `runAsNonRoot` will be set to true in the Pod security context, and `runAsUser` will be set to non-zero in the container security context, and so on.

suhasgumma:
Thank you @Daniel-GrunbergerCA . `ContainerSecurityContext` takes precedence. Only if the values are not set, we should check in `PodSecurityContext`.  I will implement the control according to this logic and get back to you.

suhasgumma:
Hi @Daniel-GrunbergerCA, this is what I have done:

* Check if `allowPreviligeEscalation is set to false` in `ContainerSecurityContext`. Only if it is not set, check for it in `PodSecurityContext`. The control fails if it is not set to false.

* Check either `runAsNonRoot` is set to `true` or `runAsUser` is `non-zero`. Only if they are not set in `ContainerSecurityContext`, fetch their values from `PodSecurityContext` if they are set there.

Daniel-GrunbergerCA (approved):
Good job! Sorry for the delay :)

matthyx:
@slashben can you have a look, Daniel did approve in the past, but a second review before merging would be good

snowhanse:
@matthyx I came across this as I was looking at a similar problem, but wonder if it's not possible to simplify the validations to a single expression by using `variables` to standardise the podSpec, something like this:

```yaml
  variables:
  # Extract podSpec from Pod, Deployment, ReplicaSet, DaemonSet, StatefulSet, Job, CronJob
  # CronJob has podSpec in object.spec.jobTemplate.spec
  - name: cronjobPodSpec
    expression: |
      object.kind == 'CronJob' ? object.spec.jobTemplate.spec : null
  # Pod is found directly in object.spec and all reminding kinds is found in object.spec.template.spec
  - name: otherPodSpec
    expression: |
      object.kind == 'Pod' ? object.spec : object.spec.template.spec
  # Pass the the correct podSpec based on the object kind
  - name: podSpec
    expression: |
      object.kind == 'CronJob' ? variables.cronjobPodSpec : variables.otherPodSpec

  validations:
    - expression: > 
        variables.podSpec.containers.all(container,
          (
            (
              has(container.securityContext) && 
              has(container.securityContext.allowPrivilegeEscalation) && 
              container.securityContext.allowPrivilegeEscalation == false
            ) ||
            (
              (
                !has(container.securityContext) || !has(container.securityContext.allowPrivilegeEscalation)
              ) &&
              (
                has(variables.podSpec.securityContext) && 
                has(variables.podSpec.securityContext.allowPrivilegeEscalation) && 
                variables.podSpec.securityContext.allowPrivilegeEscalation == false
              )
            )
          ) &&
          (
            (
              (
                has(container.securityContext) && 
                has(container.securityContext.runAsNonRoot) && 
                container.securityContext.runAsNonRoot == true
              ) ||
              (
                (
                  !has(container.securityContext) || !has(container.securityContext.runAsNonRoot)
                ) &&
                (
                  has(variables.podSpec.securityContext) && 
                  has(variables.podSpec.securityContext.runAsNonRoot) && 
                  variables.podSpec.securityContext.runAsNonRoot == true
                )
              )  
            ) ||
            (
              (
                has(container.securityContext) && 
                has(container.securityContext.runAsUser) && 
                container.securityContext.runAsUser != 0
              ) ||
              (
                (
                  !has(container.securityContext) || !has(container.securityContext.runAsUser)
                ) &&
                (
                  has(variables.podSpec.securityContext) && 
                  has(variables.podSpec.securityContext.runAsUser) && 
                  variables.podSpec.securityContext.runAsUser != 0
                )
              )  
            )
          )
        )
      message: "Workload contains container/s which have the capability to run as root! (see more at https://hub.armosec.io/docs/c-0013)"
```

slashben (approved):

PR #15, opened 2023-01-12
ValidatingAdmissionPolicy for C-0057

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/rule-privileged-container/raw.rego


slashben:
> Hi @slashben, I have added two extra test files because I could not create a list or add elements to the list. We can remove these and adjust the test cases when we figure out how to deal with the list items scenario.

Ok, I will check this (remind me please next week)

slashben (approved):

PR #21, opened 2023-01-15
ValidatingAdmissionPolicy for C-0076

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/label-usage-for-resources/raw.rego

### DIFF
--- a/controls/C-0076/policy.yaml
+++ b/controls/C-0076/policy.yaml
@@ -1,0 +1,63 @@
+apiVersion: admissionregistration.k8s.io/v1alpha1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "kubescape-c-0076-deny-resources-without-configured-list-of-labels-not-set"
+spec:
+  failurePolicy: Fail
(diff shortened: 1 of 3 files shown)


suhasgumma:
Hi @slashben, I could not integrate the following test cases as we are `using labels to bind` the `ValidatingAdmissionPolicy`
* Pod/ Deployment without labels is denied.
* Pod/Deployment with labels only from the configured list is allowed.

I have tested these cases manually and they are working as expected.

slashben:
I will look into this

slashben:
@suhasgumma , if we (for this control) create a test where the policy is bound via namespace (and not Pod/Deployment), would it work?

suhasgumma:
> @suhasgumma , if we (for this control) create a test where the policy is bound via namespace (and not Pod/Deployment), would it work?

Hi @slashben , It would work if we bind them using a namespace. I will update it.

slashben (approved):

PR #22, opened 2023-01-15
ValidatingAdmissionPolicy for  C-0077

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/K8s%20common%20labels%20usage/raw.rego

### DIFF
--- a/controls/C-0077/policy.yaml
+++ b/controls/C-0077/policy.yaml
@@ -1,0 +1,61 @@
+apiVersion: admissionregistration.k8s.io/v1alpha1
+kind: ValidatingAdmissionPolicy
+metadata:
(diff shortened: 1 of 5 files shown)


suhasgumma:
* Hi @slashben, I had to create new test files because adding labels to existing YAML with dots in the middle(`app.kubernetes.io/name`) is not possible.

* After all the controls are done, maybe we can figure out how to deal with these scenarios and reduce the number of test files.

slashben:
> * Hi @slashben, I had to create new test files because adding labels to existing YAML with dots in the middle(`app.kubernetes.io/name`) is not possible.
> * After all the controls are done, maybe we can figure out how to deal with these scenarios and reduce the number of test files.

Agreed. Let's meet and discuss.

slashben (approved):

PR #25, opened 2023-01-16
ValidatingAdmissionPolicy for C-0073

suhasgumma:
### Control Rego: https://github.com/kubescape/regolibrary/blob/master/rules/naked-pods/raw.rego

### DIFF
--- a/controls/C-0073/policy.yaml
+++ b/controls/C-0073/policy.yaml
@@ -1,0 +1,15 @@
+apiVersion: admissionregistration.k8s.io/v1alpha1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "kubescape-c-0073-deny-naked-pods"
+spec:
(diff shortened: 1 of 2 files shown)


suhasgumma:
Hi @slashben, I could not write solid tests for now because it requires a pre-existing Deployment/ReplicaSer and this needs changes in the `test script`. I am thinking of gathering all the changes needed and making changes optimally after finishing all the controls.

slashben:
> Hi @slashben, I could not write solid tests for now because it requires a pre-existing Deployment/ReplicaSer and this needs changes in the `test script`. I am thinking of gathering all the changes needed and making changes optimally after finishing all the controls.

I am for meeting and solving these test issues! 

But I am not sure that it is the case in these control tests. 

Here I'd check that a naked Pod is blocked, but deployment can start Pods.

slashben (approved):

PR #26, opened 2023-01-16
ValidatingAdmissionPolicy for  C-0012

suhasgumma:
### Control Rego: 
* https://github.com/kubescape/regolibrary/blob/master/rules/rule-credentials-in-env-var/raw.rego
* https://github.com/kubescape/regolibrary/blob/master/rules/rule-credentials-configmap/raw.rego


controls/C-0012/policy.yaml
```diff
@@ -0,0 +1,60 @@
...
+        ))
+      message: "CronJob has one or more containers with sensitive information in environment variables! (see more at https://hub.armosec.io/docs/c-0012)"
+    - expression: > 
+        object.kind != 'ConfigMap' || object.data.all(key,
+        (
+          (
+            !params.settings.sensitiveKeyNames.exists(sensitiveKey, key.lowerAscii().contains(sensitiveKey)) &&
+            !params.settings.sensitiveValues.exists(sensitiveVal, key.lowerAscii().matches(sensitiveVal))
+          ) ||
+          object.data.key == "" ||
```
suhasgumma, line 57: Hi @slashben, I want to access the variable `key` that I have defined in `object.data.all`, but CEL is taking it literally and searching for `"key"`. 

I searched for a way to access map values inside `all` macros but to no avail. That is the only thing blocking the completion of this control.

slashben: This is not good:
```
object.data.key == ""
```
I think you meant
```
key == ""
```

controls/C-0012/policy.yaml
```diff
@@ -0,0 +1,60 @@
...
+          params.settings.sensitiveValuesAllowed.exists(allowedVal, object.data.key == allowedVal)
```
slashben, line 58: @suhasgumma , the same in this line. This is what you meant here I think
```
params.settings.sensitiveValuesAllowed.exists(allowedVal, key == allowedVal)
```

controls/C-0012/policy.yaml
suhasgumma: * Hi Ben, I have been looking to access the value in the map corresponding to that key. 
* I have to use `[]` to do so. I updated and it is working as expected.

slashben:
@suhasgumma , I think there might be an issue with the way you handle environment variables. (correct me if I am wrong)

An environment variable can have a value defined in the pod spec, or it can get its value from a secret or a configmap.

For example, I have an environment variable "AWS_SECRET_KEY" defined in the pod spec. If it has an immediate value specified  in the Pod spec it should fail the control/policy. If it is mounted from a Kubernetes secret it is *good practice*!
See here: https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables

suhasgumma:
> An environment variable can have a value defined in the pod spec, or it can get its value from a secret or a configmap.

* Yes @slashben, you are right. But, the policy will not fail if the value is mounted by a secret or configMap. We are checking for them indirectly. 
* If the `value` is not set or set to a `non-empty string`, only then `secretKeyRef` or `configMapKeyRef` can be defined. 

*  If the `value` is not set or set to a `non-empty string`, the control will pass regardless. So, we don't need to check if there is `secretKeyRef` or `configMapKeyRef` set.

slashben:
You can add comment to the policy itself, no?

suhasgumma:
> You can add comment to the policy itself, no?

Yeah, I will do that and clearly specify it in the documentation.

matthyx:
@suhasgumma can you rebase and fix conflicts?

PR #51, opened 2023-10-02
Removing deny from policy names

amitschendel:


matthyx:
@amitschendel do you still need this?
If so pls rebase.
If not, pls close.

amitschendel:
@matthyx 
Honestly I don't remember...
Do you think we need it?

slashben:
I don't think we need it

PR #52, opened 2023-12-12
fix: test directory path was missing the dir prefix

Tyrael:
You can see from the Actions output that no test is being executed since adding the bogus if directory check:
Compare the "Running all control policy tests" section of the latest:
https://github.com/kubescape/cel-admission-library/actions/runs/6380302477/job/17314432072
and the last "real" run:
https://github.com/kubescape/cel-admission-library/actions/runs/6239124590/job/16936381023

### DIFF
--- a/scripts/run-all-control-tests.sh
+++ b/scripts/run-all-control-tests.sh
@@ -29,7 +29,7 @@ fi
 result=0
 # Run all control tests
 for control in $(ls controls); do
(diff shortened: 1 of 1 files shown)


slashben (approved):

slashben:
Thanks @Tyrael , I don't know how this went under the radar before 😮 
(koszi a segitseget 😉 )

PR #63, opened 2025-03-10
Add Controls C-0268, C-0269, C-0270, C-0271 to allow for more fine granular testing of cpu/mem limits/requests being set

poehlerflorian:
Currently it is only possible to check for cpu/memory requests **and** limits. Since it is not recommended to set cpu limits when cpu requests are correctly set (see [this post e.g.](https://home.robusta.dev/blog/stop-using-cpu-limits)) the controls C-0004 and C-0050 are being split up into C-0268 to C-0271 like it has already been done in the regolibrary (https://github.com/kubescape/regolibrary/pull/594).

Also the control C-0050 contained a bug where cpu requests were compared to the configured max/min cpu limit.


slashben:
Thanks @poehlerflorian for the contribution, I will look into it a bit later!

slashben (approved):

PR #68, opened 2025-04-14
Remove whitespaces at EOL

poehlerflorian:


matthyx (approved):
thanks @poehlerflorian maybe next time check that files also end with newlines (setting 'Files: Insert Final Newline' in the 'Workspace Settings' and/or 'User Settings' tab(s))

PR #79, opened 2025-10-24
fix: Allow usage of floats on C-0004,  C-0269 and C-0271

wangwillian0:
The current rule doesn't work if the resource uses float quantities on memory, like `1.5Gi`. I rewrote the rules based on https://github.com/kubescape/cel-admission-library/pull/78.

### DIFF
--- a/configuration/basic-control-configuration.yaml
+++ b/configuration/basic-control-configuration.yaml
@@ -45,10 +45,10 @@ settings:
   - 5
   maxHighVulnerabilities:
   - 10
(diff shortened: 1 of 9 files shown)


slashben:
Hey @wangwillian0 , thanks for the contribution! Can you fix the DCO problem? Otherwise it looks good and I will merge it

slashben (approved):

slashben:
Great job @wangwillian0 !

PR #81, opened 2026-05-03
feat: add CEL policies for C-0210 (seccomp) and C-0198 (root containers)

manmathbh:
Adds two new ValidatingAdmissionPolicy controls to the CEL admission library.

## C-0210 — Deny seccomp profile Unconfined
- Requires `securityContext.seccompProfile.type` to be set and not `Unconfined`
- Supports Pod, Deployment, ReplicaSet, DaemonSet, StatefulSet, Job, CronJob
- 6 test cases covering RuntimeDefault/Localhost (pass) and Unconfined/missing (fail)

## C-0198 — Deny root containers
- Denies containers that explicitly set `runAsUser: 0`
- Allows `runAsNonRoot: true` or `runAsUser != 0` as alternatives
- Complements existing C-0013 (runAsNonRoot check)
- 6 test cases covering valid configurations and blocked root user

## Related
- LFX Mentorship contributions for kubescape/kubescape#2001

## Summary by CodeRabbit

* **New Features**
  * Added two admission policies: one to deny containers running as root (UID 0) and one to deny Unconfined seccomp profiles across Pods and supported workloads.
* **Tests**
  * Added test suites covering pod and workload scenarios for runAsUser/runAsNonRoot and seccompProfile enforcement (multiple pass/fail cases).
* **Documentation**
  * Added documentation pages and updated the library listing for both policies.
* **Chores**
  * Updated control manifests to replace one policy entry with the new seccomp policy.


manmathbh:
@matthyx PTAL — part of my LFX mentorship contributions for kubescape/kubescape#2001. Thanks!

manmathbh:
PTAL @slashben

controls/C-0198/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments","replicasets","daemonsets","statefulsets"]
+    - apiGroups:   ["batch"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["jobs","cronjobs"]
+  validations:
+    - expression: "object.kind != 'Pod' || object.spec.containers.all(container, !has(container.securityContext) || !has(container.securityContext.runAsUser) || container.securityContext.runAsUser != 0 || (has(container.securityContext.runAsNonRoot) && container.securityContext.runAsNonRoot == true))"
```
matthyx: This only reads `container.securityContext.runAsUser`, so `spec.securityContext.runAsUser: 0` (and the template/jobTemplate equivalents in the branches below) is never considered. Containers inherit that UID, which means workloads can still run as root and pass this policy.

controls/C-0198/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+      message: "Pods must not explicitly run containers as root (UID 0) (see more at https://kubescape.io/docs/controls/c-0198/)"
+    - expression: "['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job'].all(kind, object.kind != kind) || object.spec.template.spec.containers.all(container, !has(container.securityContext) || !has(container.securityContext.runAsUser) || container.securityContext.runAsUser != 0 || (has(container.securityContext.runAsNonRoot) && container.securityContext.runAsNonRoot == true))"
```
matthyx: The `... || (has(runAsNonRoot) && runAsNonRoot == true)` clause lets an explicit `runAsUser: 0` pass whenever the same container also sets `runAsNonRoot: true`. That masks the root UID instead of rejecting it.

controls/C-0210/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments","replicasets","daemonsets","statefulsets"]
+    - apiGroups:   ["batch"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["jobs","cronjobs"]
+  validations:
+    - expression: "object.kind != 'Pod' || object.spec.containers.all(container, has(container.securityContext) && has(container.securityContext.seccompProfile) && has(container.securityContext.seccompProfile.type) && container.securityContext.seccompProfile.type != 'Unconfined')"
```
matthyx: This requires every container to set its own `securityContext.seccompProfile.type`, but Kubernetes also allows `spec.securityContext.seccompProfile.type` (and the template/jobTemplate equivalents). A valid pod-level `RuntimeDefault` is currently rejected here even though containers inherit it.

controls/C-0210/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+      message: "Pods must define a seccomp profile other than Unconfined (see more at https://kubescape.io/docs/controls/c-0210/)"
+    - expression: "['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job'].all(kind, object.kind != kind) || object.spec.template.spec.containers.all(container, has(container.securityContext) && has(container.securityContext.seccompProfile) && has(container.securityContext.seccompProfile.type) && container.securityContext.seccompProfile.type != 'Unconfined')"
```
matthyx: These validations iterate `containers` only. An `initContainer` can still set `seccompProfile.type: Unconfined` and be admitted, so the control has a straightforward bypass.

manmathbh:
Thanks for the review, @matthyx! All four issues are sorted:

Pod inheritance: Added a ternary fallback (container -> pod -> safe default) so denials always trigger.
runAsNonRoot: Dropped the || clause; runAsUser: 0 is strictly blocked now.
Container bypass: Loops now cover containers, initContainers, and ephemeralContainers.
Refactor: Consolidated the logic to match C-0013 and updated the docs.

 PTAL

controls/C-0198/policy.yaml
```diff
@@ -0,0 +1,55 @@
...
+        object.kind == 'Pod'
+          ? object.spec.containers + (has(object.spec.initContainers) ? object.spec.initContainers : []) + (has(object.spec.ephemeralContainers) ? object.spec.ephemeralContainers : [])
+          : object.kind in ['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job']
+            ? object.spec.template.spec.containers + (has(object.spec.template.spec.initContainers) ? object.spec.template.spec.initContainers : []) + (has(object.spec.template.spec.ephemeralContainers) ? object.spec.template.spec.ephemeralContainers : [])
+            : object.kind == 'CronJob'
+              ? object.spec.jobTemplate.spec.template.spec.containers + (has(object.spec.jobTemplate.spec.template.spec.initContainers) ? object.spec.jobTemplate.spec.template.spec.initContainers : []) + (has(object.spec.jobTemplate.spec.template.spec.ephemeralContainers) ? object.spec.jobTemplate.spec.template.spec.ephemeralContainers : [])
+              : []
+      name: containers
+    - expression: |
+        object.kind == 'Pod'
```
matthyx, line 36: `podSecurityContext` is assigned via direct field access (`object.spec.securityContext`, template spec, etc.) without `has(...)` guards. On manifests that omit pod-level `securityContext`, CEL evaluates this as `no such key: securityContext`; with `failurePolicy: Fail`, valid allow-cases are denied instead of passing. Please default this variable to `{}` when the field is absent.

controls/C-0210/policy.yaml
```diff
@@ -0,0 +1,55 @@
...
+        object.kind == 'Pod'
+          ? object.spec.containers + (has(object.spec.initContainers) ? object.spec.initContainers : []) + (has(object.spec.ephemeralContainers) ? object.spec.ephemeralContainers : [])
+          : object.kind in ['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job']
+            ? object.spec.template.spec.containers + (has(object.spec.template.spec.initContainers) ? object.spec.template.spec.initContainers : []) + (has(object.spec.template.spec.ephemeralContainers) ? object.spec.template.spec.ephemeralContainers : [])
+            : object.kind == 'CronJob'
+              ? object.spec.jobTemplate.spec.template.spec.containers + (has(object.spec.jobTemplate.spec.template.spec.initContainers) ? object.spec.jobTemplate.spec.template.spec.initContainers : []) + (has(object.spec.jobTemplate.spec.template.spec.ephemeralContainers) ? object.spec.jobTemplate.spec.template.spec.ephemeralContainers : [])
+              : []
+      name: containers
+    - expression: |
+        object.kind == 'Pod'
```
matthyx, line 36: Same issue here: `podSecurityContext` reads `object.spec.securityContext` / template equivalents directly, so a manifest with only container-level seccomp config errors with `no such key: securityContext`. Because `failurePolicy: Fail` is set, compliant workloads are rejected. This needs `has(...) ? ... : {}` guards around each path.

matthyx (commented):
Rechecked the latest changes. The earlier blockers around pod-level inheritance, `initContainers`, and kustomization look fixed, but there is still one correctness issue in both new policies.

`controls/C-0198/policy.yaml` and `controls/C-0210/policy.yaml` build `podSecurityContext` via direct access to `object.spec.securityContext` / template equivalents without `has(...)` guards. On manifests that omit pod-level `securityContext`, CEL evaluates with `no such key: securityContext`; with `failurePolicy: Fail`, that turns intended allow-cases into admission denials.

This still needs `has(...) ? ... : {}` guards around each pod/template/jobTemplate `securityContext` path.

manmathbh:
@matthyx  , the `podSecurityContext` variable was accessing `object.spec.securityContext` (and the template/jobTemplate equivalents) directly, so manifests without a pod-level `securityContext` hit no such key and got denied even when they shouldn't have. Wrapped each path with has(...) ? ... : {} so it falls back to {} instead of erroring. Fixed in both C-0198 and C-0210.
PTAL~

controls/C-0198/policy.yaml
```diff
@@ -0,0 +1,55 @@
...
+  annotations:
+    controlUrl: "https://kubescape.io/docs/controls/c-0198/"
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   [""]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["pods"]
```
matthyx: This policy now evaluates `ephemeralContainers`, but `matchConstraints` still only matches `resources: ["pods"]`. Injecting a debug container goes through the `pods/ephemeralcontainers` subresource, and `pods` alone does not match that path, so a root ephemeral container can still be added without this policy running. Please add `pods/ephemeralcontainers` (or `pods/*`) if ephemeral containers are meant to be covered.

controls/C-0210/policy.yaml
```diff
@@ -0,0 +1,55 @@
...
+  annotations:
+    controlUrl: "https://kubescape.io/docs/controls/c-0210/"
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   [""]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["pods"]
```
matthyx: Same bypass here: the CEL logic includes `ephemeralContainers`, but the policy only matches `resources: ["pods"]`. Ephemeral debug container injection is an update to the `pods/ephemeralcontainers` subresource, so an `Unconfined` ephemeral container can still be attached without this policy being evaluated. Please match `pods/ephemeralcontainers` (or `pods/*`) as well.

manmathbh:
@matthyx , ephemeral containers use the `pods/ephemeralcontainers` subresource, not regular pod updates, so the policy was never firing on that path even though the CEL handles it. Added `pods/ephemeralcontainers` to the resource list in both policies.
PTAL~

matthyx (approved):
all good thanks

PR #82, opened 2026-05-03
feat: add CEL policies for C-0199, C-0200, and C-0201 (capability admission controls)

manmathbh:
Three new ValidatingAdmissionPolicy controls targeting Kubernetes Pod Security Admission capability restrictions:

- **C-0199 (Deny NET_RAW capability)**: Blocks admission of containers with the potentially dangerous NET_RAW capability. The NET_RAW capability allows containers to open raw sockets, which can be misused for network attacks.

- **C-0200 (Deny added capabilities)**: Blocks admission of containers with capabilities assigned beyond the default container runtime set. Enforces the principle of least privilege by requiring containers to run with only the default capabilities.

- **C-0201 (Deny capabilities assigned)**: Blocks admission of containers with any capabilities explicitly assigned. Aligns with the PSA restricted standard requiring that all capabilities must be dropped.

Each control includes:
- CEL expressions for Pod, workload (Deployment/ReplicaSet/DaemonSet/StatefulSet/Job), and CronJob resources
- Test cases in `tests.json` covering pass/fail scenarios across different resource types

These build on the existing capability-related controls (C-0046 for insecure capabilities) by providing finer-grained admission policies for specific capability profiles.

## Summary by CodeRabbit

* **New Features**
  * Added three admission policies enforcing capability rules: deny NET_RAW, deny added capabilities, and require dropping ALL across Pods and common workload types.

* **Tests**
  * Added extensive test suites covering Pods, Deployments, Jobs/CronJobs, init and ephemeral containers with pass/fail scenarios.

* **Documentation**
  * Added policy documentation and updated the Library items table with entries and links for each control.

* **Chores**
  * Registered new policies in control manifests and added a test resource for drop-ALL scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/cel-admission-library/pull/82)


controls/C-0199/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+      message: "Container has the NET_RAW capability which is potentially dangerous! (see more at https://kubescape.io/docs/controls/c-0199/)"
+    - expression: "['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job'].all(kind, object.kind != kind) || object.spec.template.spec.containers.all(container, !has(container.securityContext) || !has(container.securityContext.capabilities) || !has(container.securityContext.capabilities.add) || container.securityContext.capabilities.add.all(capability, capability != 'NET_RAW'))"
+      message: "Workload has a container with the NET_RAW capability which is potentially dangerous! (see more at https://kubescape.io/docs/controls/c-0199/)"
+    - expression: "object.kind != 'CronJob' || object.spec.jobTemplate.spec.template.spec.containers.all(container, !has(container.securityContext) || !has(container.securityContext.capabilities) || !has(container.securityContext.capabilities.add) || container.securityContext.capabilities.add.all(capability, capability != 'NET_RAW'))"
```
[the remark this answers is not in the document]
slashben: @manmathbh you should check init and ephemeral containers

slashben:
@manmathbh there are two outstanding issue:
1. coverage in initContainers (potentialy ephemeral containers as well)
2. Strategy to handle `drop` as opposed to `add`

First is not that complex, but the second is complicated. I am tilting to enforce `drop` and not rely on the container runtime defaults

manmathbh:
> @manmathbh there are two outstanding issue:
> 
> 1. coverage in initContainers (potentialy ephemeral containers as well)
> 2. Strategy to handle `drop` as opposed to `add`
> 
> First is not that complex, but the second is complicated. I am tilting to enforce `drop` and not rely on the container runtime defaults

 @slashben both issues are addressed:

 1. initContainers / ephemeralContainers coverage — All three controls now iterate over the concatenated list containers + initContainers + ephemeralContainers using CEL list concatenation with has() guards. No container type is skipped. Tests added for initContainer scenarios.

 2. drop vs add strategy — Enforcing drop as suggested, no reliance on runtime defaults:

 | Control | Before (all three were variants of checking .add) | After |
 |---------|-----------------------------------------------------|-------|
 | C-0199 | add doesn't contain NET_RAW | Unchanged — targeted capability ban, .add is the right layer |
 | C-0200 | size(add) == 0 | Now requires drop: ["ALL"] — PSA restricted baseline, no runtime defaults |
 | C-0201 | Identical to C-0200 | Now requires drop: ["ALL"] AND size(add) == 0 — strictest posture, drops everything + adds nothing |

 New test fixture test-resources/pod-drop-all.yaml added. PTAL. Ready for another look.

slashben (commented):
The CEL expressions work correctly, but for future contributions please use YAML folded style (`>`) with line breaks — the single-line format is hard to review and maintain. You can also look at `C-0013/policy.yaml` for an example of using `spec.variables` to extract the container lists once, which avoids repeating the init/ephemeral container gathering logic three times per policy.

Not blocking merge on this, just something to keep in mind for next time.

---

Could you add the doc files under `docs/policies-based-on-kubescape-controls/` for C-0199, C-0200, and C-0201? The README links to them but they're not included in the PR, so the links will be broken. You can use any existing doc file in that directory as a template.

---

Could you add a few more test cases? Specifically:
- An **ephemeralContainers** test for each control (the CEL handles them but there's no test covering that path)
- A **positive initContainer test** for C-0200 — there's a test for an initContainer *lacking* `drop: ALL` being denied, but no test confirming a pod with both main + init containers dropping ALL is allowed

slashben:
Great work, we are nearly there! Please fix the last 3 things (missing doc files, ephemeralContainers tests, and the positive initContainer test for C-0200) and I will merge.

manmathbh:
> Great work, we are nearly there! Please fix the last 3 things (missing doc files, ephemeralContainers tests, and the positive initContainer test for C-0200) and I will merge.

@slashben all three done:

 - Doc files added under docs/policies-based-on-kubescape-controls/ for C-0199, C-0200, C-0201
 - EphemeralContainers tests added for all three controls
 - Positive initContainer test for C-0200, pod with both main + init containers dropping ALL → pass

 PTAL

controls/C-0199/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+  annotations:
+    controlUrl: "https://kubescape.io/docs/controls/c-0199/"
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   [""]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["pods"]
```
slashben: One thing I missed on the earlier pass: ephemeralContainers are normally added via the `pods/ephemeralcontainers` subresource (that's what `kubectl debug` hits), not via PUT/PATCH on `pods`. With the current matchConstraints the ephemeralContainers branch of the CEL won't fire on the typical add-ephemeral path. Could you add a rule for `pods/ephemeralcontainers` so this is actually covered? Same applies to C-0200 and C-0201.

slashben (commented):
Great work, thanks for being so responsive. Two things before merge:

1. EphemeralContainers coverage gap, left an inline comment on `controls/C-0199/policy.yaml`. Same change needed on C-0200 and C-0201 policy.yaml.
2. There's a merge conflict with `main` now. Could you rebase (or merge `main` into your branch) once the subresource fix is in?

Once that lands I'll merge.

manmathbh:
@slashben 
should be good to go
PTAL

manmathbh:
@slashben gentle ping
both items from your last review are in, all checks green.
PTAL.

manmathbh:
@slashben, rebased onto main,
added the Job and positive ephemeralContainer tests,
and fixed the C-0200 description to say "require drop all" so it matches what the expression actually checks.
PTAL

manmathbh:
@slashben all checks gree, 
thanks for the detailed reviews on this one,
learned a lot from the back and forth.
ready whenever you are
PTAL

matthyx (changes requested):
One additional blocker beyond the existing `pods/ephemeralcontainers` discussion: the new negative tests currently pass on **any** non-zero `kubectl` exit. In `scripts/run-control-tests.py`, `expected: "fail"` is treated as success without checking that the rejection actually came from the policy under test. That means these cases can go green because of an unrelated schema/admission error instead of a VAP denial.

This is especially risky for the new ephemeralContainer cases, because adding `spec.ephemeralContainers` via a normal Pod apply is not the same as exercising the `pods/ephemeralcontainers` subresource. Please make the harness assert on the expected policy rejection (message/policy name) or otherwise distinguish the intended VAP denial before merge.

matthyx:
Re-checked after 58f07dd.

The earlier false-positive in the negative test harness is fixed now: `scripts/run-control-tests.py` no longer treats any non-zero `kubectl` exit as success, and instead requires stderr to mention the expected policy name.

One blocker still remains though: the new ephemeral-container cases in `controls/C-0199/tests.json`, `controls/C-0200/tests.json`, and `controls/C-0201/tests.json` still inject `spec.ephemeralContainers` into a Pod manifest and run `kubectl --dry-run=server apply -f <pod>`. That exercises normal Pod create/update, not the `pods/ephemeralcontainers` subresource that `kubectl debug` / ephemeral-container admission actually uses, so these cases can still fail for the wrong reason before the VAP under test is what rejects them.

I think this should either:
1. drop those three ephemeral-container cases for now, or
2. teach the test harness to exercise the real `pods/ephemeralcontainers` path and assert the VAP denial there.

Everything else I re-checked looks good at blocker level.

manmathbh:
@matthyx, agreed that checking only the exit code was too loose. Fixed in the latest commit. Changed `subprocess.check_call `to subprocess.run with `capture_output=True`, and now for expected: "fail" it verifies the policy name appears in stderr before counting it as a pass. Should prevent false positives from unrelated kubectl errors.
PTAL.

manmathbh:
> Re-checked after [58f07dd](https://github.com/kubescape/cel-admission-library/commit/58f07dde29dc2c6914760715b2e74544d8a0308a).
> 
> The earlier false-positive in the negative test harness is fixed now: `scripts/run-control-tests.py` no longer treats any non-zero `kubectl` exit as success, and instead requires stderr to mention the expected policy name.
> 
> One blocker still remains though: the new ephemeral-container cases in `controls/C-0199/tests.json`, `controls/C-0200/tests.json`, and `controls/C-0201/tests.json` still inject `spec.ephemeralContainers` into a Pod manifest and run `kubectl --dry-run=server apply -f <pod>`. That exercises normal Pod create/update, not the `pods/ephemeralcontainers` subresource that `kubectl debug` / ephemeral-container admission actually uses, so these cases can still fail for the wrong reason before the VAP under test is what rejects them.
> 
> I think this should either:
> 
> 1. drop those three ephemeral-container cases for now, or
> 2. teach the test harness to exercise the real `pods/ephemeralcontainers` path and assert the VAP denial there.
> 
> Everything else I re-checked looks good at blocker level.

@matthyx Makes total sense. Those tests can't exercise the subresource path properly without a harness rewrite.
I went with **Option 1** to keep the scope focused, I have dropped the three ephemeral container cases from `C-0199`, `C-0200`, and `C-0201` in the latest commit. The `initContainer` tests still cover the CEL list iteration, and the `pods/ephemeralcontainers `subresource is still safely in `matchConstraints` for production use.
PTAL!

controls/C-0200/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments","replicasets","daemonsets","statefulsets"]
+    - apiGroups:   ["batch"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["jobs","cronjobs"]
+  validations:
+    - expression: "object.kind != 'Pod' || (object.spec.containers + (has(object.spec.initContainers) ? object.spec.initContainers : []) + (has(object.spec.ephemeralContainers) ? object.spec.ephemeralContainers : [])).all(container, has(container.securityContext) && has(container.securityContext.capabilities) && has(container.securityContext.capabilities.drop) && container.securityContext.capabilities.drop.exists(d, d == 'ALL'))"
```
matthyx: C-0200 still looks mapped to the wrong behavior. Kubescape C-0200 is about denying capabilities added beyond the default runtime set (`securityContext.capabilities.add` should be absent/empty). This expression instead requires every container to set `capabilities.drop=["ALL"]`, which rejects workloads that add no extra capabilities at all and overlaps with the stricter C-0201 posture. The earlier test-harness/subresource blockers look fixed, but this control still does not match the upstream C-0200 semantics.

matthyx (changes requested):
Rechecked the current head and two semantic blockers still remain:

1. `controls/C-0200/{policy.yaml,tests.json}` still implements a "drop ALL capabilities" rule. Official Kubescape C-0200 is the control for denying capabilities beyond the default runtime set, so this should validate `capabilities.add`, not require `capabilities.drop: ["ALL"]`. As written, C-0200 overlaps with C-0201 instead of implementing its own behavior.

2. `controls/C-0199/{policy.yaml,tests.json}` still only rejects explicit `capabilities.add: ["NET_RAW"]`. The control is about not permitting workloads with the `NET_RAW` capability, and runtimes commonly include `NET_RAW` in the default capability set, so workloads with no capabilities stanza still pass here when they should have to drop `NET_RAW` (or `ALL`).

`C-0201` looks fine now.

manmathbh:
> Rechecked the current head and two semantic blockers still remain:
> 
> 1. `controls/C-0200/{policy.yaml,tests.json}` still implements a "drop ALL capabilities" rule. Official Kubescape C-0200 is the control for denying capabilities beyond the default runtime set, so this should validate `capabilities.add`, not require `capabilities.drop: ["ALL"]`. As written, C-0200 overlaps with C-0201 instead of implementing its own behavior.
> 2. `controls/C-0199/{policy.yaml,tests.json}` still only rejects explicit `capabilities.add: ["NET_RAW"]`. The control is about not permitting workloads with the `NET_RAW` capability, and runtimes commonly include `NET_RAW` in the default capability set, so workloads with no capabilities stanza still pass here when they should have to drop `NET_RAW` (or `ALL`).
> 
> `C-0201` looks fine now.

@matthyx both semantic blockers are fixed now.
**C-0199**, changed from checking `capabilities.add` to requiring `capabilities.drop` contains NET_RAW or ALL, so workloads with no capabilities stanza are correctly rejected.
**C-0200**, reverted to checking `capabilities.add` is absent, matching the upstream Kubescape **C-0200** semantics (deny capabilities beyond the default runtime set).
No longer overlaps with **C-0201**.
Updated both `tests.json` suites to match.
**C-0201** remains unchanged.
PTAL!

controls/C-0199/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments","replicasets","daemonsets","statefulsets"]
+    - apiGroups:   ["batch"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["jobs","cronjobs"]
+  validations:
+    - expression: "object.kind != 'Pod' || (object.spec.containers + (has(object.spec.initContainers) ? object.spec.initContainers : []) + (has(object.spec.ephemeralContainers) ? object.spec.ephemeralContainers : [])).all(container, has(container.securityContext) && has(container.securityContext.capabilities) && has(container.securityContext.capabilities.drop) && container.securityContext.capabilities.drop.exists(c, c == 'NET_RAW' || c == 'ALL'))"
```
matthyx: This still doesn't fully enforce **"deny NET_RAW capability"**.

Right now the check only requires `capabilities.drop` to contain `NET_RAW` or `ALL`, but it never also rejects `capabilities.add: ["NET_RAW"]`. A workload can therefore satisfy this policy with `drop: ["ALL"]` and still add `NET_RAW` back, which leaves the forbidden capability effective.

Can we also require `capabilities.add` to be absent (or at least not contain `NET_RAW`) in all three expressions here, and add a regression test for the `drop: ["ALL"]` + `add: ["NET_RAW"]` case?

manmathbh: @matthyx good catch, drop: ["`ALL`"] + add: ["`NET_RAW`"] re-enables the forbidden capability.
Fixed all three expressions to also require add doesn't contain `NET_RAW`, and added a regression test for that edge case. Should be fully covered now.
PTAL.

controls/C-0199/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments","replicasets","daemonsets","statefulsets"]
+    - apiGroups:   ["batch"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["jobs","cronjobs"]
+  validations:
+    - expression: "object.kind != 'Pod' || (object.spec.containers + (has(object.spec.initContainers) ? object.spec.initContainers : []) + (has(object.spec.ephemeralContainers) ? object.spec.ephemeralContainers : [])).all(container, has(container.securityContext) && has(container.securityContext.capabilities) && has(container.securityContext.capabilities.drop) && container.securityContext.capabilities.drop.exists(c, c == 'NET_RAW' || c == 'ALL') && (!has(container.securityContext.capabilities.add) || container.securityContext.capabilities.add.all(cap, cap != 'NET_RAW')))"
```
matthyx: One blocker still remains in `C-0199` here: Kubernetes allows `securityContext.capabilities.add: ["ALL"]`.

With the current predicate `(!has(container.securityContext.capabilities.add) || container.securityContext.capabilities.add.all(cap, cap != 'NET_RAW'))`, a workload can still satisfy the policy with `drop: ["ALL"]` and then `add: ["ALL"]`, which re-enables `NET_RAW` (and every other capability).

Can we also treat `ALL` as forbidden in `capabilities.add` for all three expressions, and add a regression test for the `drop: ["ALL"]` + `add: ["ALL"]` case?

matthyx (approved):
Rechecked the current head. The prior C-0199 and C-0200 semantic blockers are fixed, including the C-0199 drop:["ALL"] + add:["ALL"] loophole. No remaining blocking issues from my side.

PR #84, opened 2026-05-06
feat: add C-0280 - deny access to certificatesigningrequests/approval sub-resource

manmathbh:
## Summary
- Implements C-0280 control (CIS Kubernetes Benchmark 5.1.11): Minimize access to the approval sub-resource of certificatesigningrequests objects
- Blocks creation/update of Role and ClusterRole resources that grant access to `certificatesigningrequests/approval` or `certificatesigningrequests/*`
- First VAP policy in this library targeting `rbac.authorization.k8s.io` API group

## Changes
- **New control:** `controls/C-0280/` (policy.yaml + tests.json)
- **Test templates:** 4 CSR-approval variant YAMLs + 2 base RBAC templates in `test-resources/`
- **Documentation:** `docs/policies-based-on-kubescape-controls/kubescape-c-0280-deny-access-to-csr-approval-subresource.md`
- **Registration:** Added to `controls/kustomization.yaml` and `README.md`

## Tests
- 6 test cases: 2 pass (clean roles), 4 fail (CSR approval / CSR subresource wildcard)
- Tests require a running K8s cluster with VAP support (validated via CI)

## Summary by CodeRabbit

* **New Features**
  * Added a new security policy (C-0280) that denies access to CSR approval sub-resources in Roles and ClusterRoles.

* **Documentation**
  * Added policy documentation describing severity and usage guidance.
  * Added policy entry to the library index.

* **Tests**
  * Added comprehensive test scenarios covering allowed/blocked RBAC cases for CSR approval.

* **Chores**
  * Registered the new policy in cluster manifests/kustomization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/cel-admission-library/pull/84)


manmathbh:
PTAL @slashben

controls/C-0280/policy.yaml
```diff
@@ -0,0 +1,26 @@
...
+      resources:   ["roles", "clusterroles"]
+  validations:
+    - expression: >
+        ['Role', 'ClusterRole'].all(kind, object.kind != kind) ||
+        object.rules.all(rule,
+          !(
+            has(rule.resources) && rule.resources.exists(r,
+              r == "certificatesigningrequests/approval" || r == "certificatesigningrequests/*")
+          )
+        )
```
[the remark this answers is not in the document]
manmathbh: added "*" and "*/*" to the resource check

manmathbh:
Found a bug!! the expression to let rolebindings through was actually blocking them (inverted logic).
Stripped those out since the policy only checks roles/clusterroles anyway. 
Much simpler now, just one expression. Rebased on main too. 
PTAL @matthyx

controls/C-0280/policy.yaml
```diff
@@ -0,0 +1,26 @@
...
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["roles", "clusterroles"]
+  validations:
+    - expression: >
+        ['Role', 'ClusterRole'].all(kind, object.kind != kind) ||
+        object.rules.all(rule,
+          !(
+            has(rule.resources) && rule.resources.exists(r,
+              r == "certificatesigningrequests/approval" || r == "certificatesigningrequests/*" || r == "*" || r == "*/*")
```
matthyx: Blocking: `resources: ["*"]` / `[*/*]` are only wildcards within the rule's `apiGroups`. As written, this will reject roles like `apiGroups: [""]`, `resources: ["*"]`, even though that only grants core-group access and does not include `certificates.k8s.io/certificatesigningrequests/approval`. Please gate the wildcard case on `apiGroups` containing `certificates.k8s.io` or `*`, and update the wildcard tests because `cluster-role.yaml` / `role.yaml` currently use the core API group.

controls/C-0280/policy.yaml
```diff
@@ -0,0 +1,26 @@
...
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["rbac.authorization.k8s.io"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["roles", "clusterroles"]
+  validations:
+    - expression: >
+        ['Role', 'ClusterRole'].all(kind, object.kind != kind) ||
+        object.rules.all(rule,
```
matthyx, line 21: Blocking: the Kubescape control definition is specifically about subjects that can `update` `certificatesigningrequests/approval`, but this expression never looks at `rule.verbs`. That makes the policy overbroad: even a read-only rule that mentions this subresource would be denied. The match needs verb gating for approval-capable verbs (at least `update`, plus wildcard verbs if you want to cover those too).

matthyx (changes requested):
Found two blocking issues in the new control implementation.

controls/C-0280/policy.yaml
manmathbh: wildcard * / */* now gates on apiGroups too.
only blocks when certs.k8s.io or * is in apiGroups, core group wildcards pass through fine.

controls/C-0280/policy.yaml
manmathbh: added a verb gate, needs update or * verb.
read-only access (e.g. get) to the subresource is harmless,
no reason to block it.

manmathbh:
@matthyx both sorted, 
wildcard now gated on certs.k8s.io/* apiGroups,
and added verb check for update/*. tests cover the new combos.
PTAL

controls/C-0280/policy.yaml
```diff
@@ -0,0 +1,31 @@
...
+          !(
+            has(rule.resources) && (
+              rule.resources.exists(r, r == "certificatesigningrequests/approval" || r == "certificatesigningrequests/*")
+              ||
+              (rule.resources.exists(r, r == "*" || r == "*/*") && has(rule.apiGroups) && rule.apiGroups.exists(g, g == "certificates.k8s.io" || g == "*"))
+            )
+            &&
+            has(rule.verbs) && rule.verbs.exists(v, v == "update" || v == "*")
```
matthyx: Still blocking: the verb predicate only matches `update` and `*`, but the CSR `approval` subresource can also be changed via `patch`. That means a Role/ClusterRole granting `certificatesigningrequests/approval` (or `certificatesigningrequests/*`) with `verbs: ["patch"]` would bypass this control. Please include `patch` in the dangerous-verb check and add a regression test for it.

matthyx (changes requested):
Re-checked the latest commit. One blocker remains.

controls/C-0280/policy.yaml
manmathbh: sure, will do

manmathbh: right, added patch alongside update and *. also added a regression tes, CSR approval with patch verb is now blocked.

matthyx (dismissed):
Re-checked the latest commit. The previously reported blockers are fixed, and I don’t see any remaining blocker-level issues in this PR.

manmathbh:
@matthyx, CI's all green now
just needs a re-stamp after the test fixture fix.

matthyx (approved):
thanks

PR #89, opened 2026-05-16
C-0050: compare CPU resources with quantity

### LINKED ISSUE #88: C-0050 should use CEL `quantity()` for CPU resources instead of `int()`
### Summary

C-0050 validates CPU requests and limits by calling `int(container.resources.requests.cpu)` and `int(container.resources.limits.cpu)`. Kubernetes CPU quantities commonly use values such as `100m`, `250m`, and `0.5`. These are valid Kubernetes resource quantities, but they are not valid integer strings.

Other CEL controls in the same repository already use `quantity()`, for example C-0268 and C-0270.

### Affected code

- `controls/C-0050/policy.yaml:30-35`
- `controls/C-0050/policy.yaml:39-44`
- `controls/C-0050/policy.yaml:48-53`

### Why this matters

C-0050 is a resource governance control. Most real workloads set CPU in millicores, for example:

```yaml
resources:
  requests:
    cpu: 250m
  limits:
    cpu: 500m
```

If the CEL policy cannot evaluate common valid Kubernetes CPU quantities, the admission policy can deny valid workloads or fail closed due to `failurePolicy: Fail`. That makes the CEL control harder to use in real clusters and makes it diverge from expected Kubernetes semantics.

### Expected behavior

The policy should compare CPU quantities using Kubernetes quantity semantics and allow values inside the configured range.

### Actual behavior

The policy attempts integer conversion with `int(...)`, which does not model Kubernetes CPU quantity strings.

### Proposed fix

- Replace `int(container.resources.requests.cpu)` and `int(container.resources.limits.cpu)` with `quantity(string(...))`.
- Compare against `quantity(string(params.settings.cpuRequestMin))`, `quantity(string(params.settings.cpuRequestMax))`, `quantity(string(params.settings.cpuLimitMin))`, and `quantity(string(params.settings.cpuLimitMax))`.
- Add test cases for `100m`, `250m`, `500m`, `1`, and out-of-range values.
- Align C-0050 with the quantity pattern already used by C-0268 and C-0270.

dakshhhhh16:
### Summary

This PR updates C-0050 to compare CPU requests and limits using Kubernetes quantity semantics. The current policy calls `int()` on CPU values, which does not support common CPU quantities such as `100m`, `250m`, or `0.5`.

### Changes

- Replace `int(container.resources.requests.cpu)` with `quantity(string(container.resources.requests.cpu))`.
- Replace `int(container.resources.limits.cpu)` with `quantity(string(container.resources.limits.cpu))`.
- Compare CPU values against `quantity(string(params.settings.*))`.
- Align C-0050 with the pattern already used in C-0268 and C-0270.
- Add tests for millicores, whole cores, decimal cores, missing resources, and out-of-range values.

### Testing

```bash
cd controls/C-0050
python ../../scripts/run-control-tests.py
```

Additional local cases to include:

```yaml
resources:
  requests:
    cpu: 250m
  limits:
    cpu: 500m
```

### Why this is safe

`quantity()` is the CEL function intended for Kubernetes resource quantity comparison. The repository already uses it for related CPU controls, so this makes C-0050 consistent with the rest of the library.

### DIFF
--- a/controls/C-0050/tests.json
+++ b/controls/C-0050/tests.json
@@ -24,6 +24,24 @@
         ]
     },
     {
+        "name": "Pod with container having cpu request and limits set in millicores is allowed",
+        "template": "pod.yaml",
+        "expected": "pass",
+        "field_change_list": [
+            "spec.containers.[0].resources.requests.cpu=250m",
+            "spec.containers.[0].resources.limits.cpu=500m"
+        ]
+    },
+    {
+        "name": "Pod with container having cpu request below the millicore minimum is blocked",
+        "template": "pod.yaml",
+        "expected": "fail",
+        "field_change_list": [
+            "spec.containers.[0].resources.requests.cpu=50m",
+            "spec.containers.[0].resources.limits.cpu=500m"
+        ]
+    },
+    {
         "name": "Pod with container having cpu request and limits set and values not in the limit is blocked",
         "template": "pod.yaml",
         "expected": "fail",
@@ -46,6 +64,24 @@
         "field_change_list": [
             "spec.template.spec.containers.[0].resources.requests.cpu=3",
             "spec.template.spec.containers.[0].resources.limits.cpu=3"
+        ]
+    },
+    {
+        "name": "Deployment with container having cpu request and limits set in millicores is allowed",
+        "template": "deployment.yaml",
+        "expected": "pass",
+        "field_change_list": [
+            "spec.template.spec.containers.[0].resources.requests.cpu=250m",
+            "spec.template.spec.containers.[0].resources.limits.cpu=500m"
+        ]
+    },
+    {
+        "name": "CronJob with container having cpu request and limits set in millicores is allowed",
+        "template": "cronjob.yaml",
+        "expected": "pass",
+        "field_change_list": [
+            "spec.jobTemplate.spec.template.spec.containers.[0].resources.requests.cpu=250m",
+            "spec.jobTemplate.spec.template.spec.containers.[0].resources.limits.cpu=500m"
         ]
     }
 ]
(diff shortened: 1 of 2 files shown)


controls/C-0050/tests.json
```diff
@@ -23,6 +23,24 @@
             "spec.containers.[0].resources.limits.cpu=3"
         ]
     },
+    {
+        "name": "Pod with container having cpu request and limits set in millicores is allowed",
+        "template": "pod.yaml",
+        "expected": "pass",
+        "field_change_list": [
+            "spec.containers.[0].resources.requests.cpu=250m",
+            "spec.containers.[0].resources.limits.cpu=500m"
```
matthyx: These new `expected: "pass"` cases are still using `limits.cpu=500m`, but `test-resources/default-control-configuration.yaml` keeps `cpuLimitMin: 1`, so the control should deny them even after switching to `quantity()`. I reproduced that locally with `python ../../scripts/run-control-tests.py` — the Pod/Deployment/CronJob millicore pass cases all fail for this reason. Could you either raise these limit fixtures to `>= 1` CPU (for example `1000m`) or override the params fixture for the millicore-success scenario?

matthyx (changes requested):
The switch from `int(...)` to `quantity(...).compareTo(...)` in `controls/C-0050/policy.yaml` looks directionally right, but the PR is still blocked by the new millicore pass expectations in `controls/C-0050/tests.json`: the default test config keeps `cpuLimitMin: 1`, so `limits.cpu=500m` is out of range and the new Pod/Deployment/CronJob `expected: pass` cases fail locally. Please either bump those limit values to at least 1 CPU (for example `1000m`) or run them against an overridden params fixture that allows `500m` limits.

matthyx (approved):
Rechecked this. The previous blocker is fixed: the millicore pass cases now use 1000m limits, which matches the default cpuLimitMin: 1, and the C-0050 control test suite passes locally.

PR #91, opened 2026-05-16
C-0001: normalize image registries before deny matching

### LINKED ISSUE #90: C-0001 should treat `nginx` and `library/nginx` as Docker Hub images when `docker.io` is forbidden
### Summary

C-0001 checks forbidden registries with `container.image.startsWith(registry)`. The default test configuration includes `docker.io` in `untrustedRegistries`, but Kubernetes image references such as `nginx` and `library/nginx` are valid Docker Hub images without an explicit `docker.io/` prefix.

As a result, a policy configured to forbid `docker.io` can still allow images that Kubernetes resolves to Docker Hub.

### Affected code

- `controls/C-0001/policy.yaml:29-34`
- `configuration/basic-control-configuration.yaml:106-110`
- `test-resources/default-control-configuration.yaml:109-113`

### Why this matters

Registry allow and deny controls are only useful if they evaluate the registry that Kubernetes will actually pull from. Without Docker Hub normalization, users can bypass a `docker.io` block by omitting the registry prefix:

```yaml
image: nginx
```

This is especially relevant to the CEL migration work because C-0078 already contains explicit logic for implicit Docker Hub images. C-0001 should use the same normalization pattern so equivalent controls behave consistently.

### Expected behavior

The policy should deny the image because `nginx` resolves to Docker Hub.

### Actual behavior

The expression only checks `container.image.startsWith('docker.io')`, so the bare image name does not match.

### Proposed fix

- Add CEL normalization for registry-less images.
- Treat images with no registry host as `docker.io`.
- Treat one-segment images such as `nginx` as `docker.io/library/nginx`.
- Avoid prefix spoofing by matching registry boundaries with `registry + '/'`, not only `startsWith(registry)`.
- Add tests for `nginx`, `library/nginx`, `docker.io/library/nginx`, `docker.io.evil.com/nginx`, and `gcr.io.evil.com/nginx`.

dakshhhhh16:
### Summary

This PR fixes C-0001 so registry-less Docker Hub images are evaluated against `docker.io` entries in `untrustedRegistries`. Today the policy only checks `container.image.startsWith(registry)`, so `image: nginx` bypasses a `docker.io` deny configuration.

Fixes #90 
### Changes

- Add CEL logic that derives the image registry before matching.
- Treat images without an explicit registry host as Docker Hub.
- Treat one-segment image names as Docker Hub library images.
- Match registry boundaries with `registry + '/'` to avoid spoofed domains such as `docker.io.evil.com`.
- Apply the same helper expression to Pod, workload template, and CronJob paths.
- Add regression tests for explicit and implicit Docker Hub images plus spoofed registry names.

### Testing

```bash
cd controls/C-0001
python ../../scripts/run-control-tests.py
```

Regression cases:

```yaml
image: nginx
```

```yaml
image: library/nginx
```

```yaml
image: docker.io.evil.com/nginx
```

### Why this is safe

The change only affects how the image registry is derived before applying the configured deny list. Users who already deny `docker.io` get the behavior Kubernetes image resolution implies. The explicit registry path continues to work.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container registry validation to reliably detect Docker Hub (including common aliases and variants), block implicit/namespace/explicit Docker Hub images when forbidden, and prevent evasion via registry-prefix spoofing. Messaging standardized to reference a "forbidden registry."

* **Refactor**
  * Merged separate resource-specific registry checks into a single, unified validation to ensure consistent enforcement across all resource types.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/cel-admission-library/pull/91?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/controls/C-0001/policy.yaml
+++ b/controls/C-0001/policy.yaml
@@ -36,11 +36,12 @@ spec:
     - expression: >
         variables.containers.all(container, params.settings.untrustedRegistries.all(registry,
         !(
-          ((registry == 'docker.io' || registry == 'docker.io/') &&
+          ((registry in ['docker.io', 'docker.io/', 'index.docker.io', 'index.docker.io/', 'registry-1.docker.io', 'registry-1.docker.io/']) &&
             (!container.image.contains('/') ||
               (!container.image.split('/')[0].contains('.') &&
                !container.image.split('/')[0].contains(':') &&
-               container.image.split('/')[0] != 'localhost'))) ||
+               container.image.split('/')[0] != 'localhost') ||
+              container.image.split('/')[0] in ['docker.io', 'index.docker.io', 'registry-1.docker.io'])) ||
           container.image.startsWith(registry.endsWith('/') ? registry : registry + '/')
         )))
       messageExpression: 'object.kind + "/" + object.metadata.name + " uses an image from a forbidden registry! (see more at https://kubescape.io/docs/controls/c-0001/)"'
--- a/controls/C-0001/policy.yaml
+++ b/controls/C-0001/policy.yaml
@@ -25,10 +25,22 @@ spec:
       apiVersions: ["v1"]
       operations:  ["CREATE", "UPDATE"]
       resources:   ["jobs","cronjobs"]
+  variables:
+    - expression: |
+        object.kind == 'Pod' ? object.spec.containers
+        : object.kind in ['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job'] ? object.spec.template.spec.containers
+        : object.kind == 'CronJob' ? object.spec.jobTemplate.spec.template.spec.containers
+        : []
+      name: containers
   validations:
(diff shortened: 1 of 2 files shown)


matthyx (changes requested):
Blocking: this still leaves a Docker Hub bypass when `docker.io` is denied. The new special-case covers implicit names like `nginx` and explicit `docker.io/...`, but `index.docker.io/library/nginx` and `registry-1.docker.io/library/nginx` both fall through because the explicit check only matches `registry + "/"`. Those hostnames are Docker Hub aliases, so a deny list containing `docker.io` can still be bypassed.

Please normalize the Docker Hub aliases before comparison (or otherwise treat them as equivalent to `docker.io`) and add regression coverage for those forms.

dakshhhhh16:
You can have a look now @matthyx

matthyx (approved):

PR #95, opened 2026-06-11
fix(C-0026): ship [Warn] binding example and rewire test for audit semantics

### LINKED ISSUE #94: C-0026 enforces audit-only Rego as a hard deny, plus library-wide variables / docs cleanup [LFX'26]
Opening this as one tracking issue for follow-up on the VAP policies that landed under #81, #82, #84, #85. Three things worth picking up before the LFX migration batch starts adding more controls. Listing them roughly in priority order.

 ### 1. C-0026 has no safe Deny semantics, and the library can't express its audit intent

PR #85 added C-0026 with the expression `object.kind != 'CronJob'`. matchConstraints binds CronJobs. So for any CronJob the validation cleanly evaluates to `false`, which is a validation failure (not an evaluation error). The binding's `validationActions` is what decides                                                                                           what happens next: `[Deny]` rejects the request, `[Audit]` or `[Warn]` records / surfaces it.
                                                                                                                                                                       
The Rego source for C-0026 in `kubescape/regolibrary` is `rule-deny-cronjobs/raw.rego`. It emits an `alert` whenever a CronJob exists. The control JSON has `baseScore: 1.0` and the test field reads "We list all CronJobs that exist in cluster for the user to approve."                                                                                              
So in Rego, this is an audit / review control, not an enforcement.                                                                                                 

This is where the library hits a wall. The repo ships only the `ValidatingAdmissionPolicy` objects, not bindings. The only binding in the tree is `test-resources/policy-binding.yaml`, which is hardcoded to `validationActions: [Deny]`, and the C-0026 test asserts `"expected": "fail"`.
So:                                                                                                                                                                
  - the policy itself cannot express "this is audit-only" , that decision lives in the binding, which the consumer writes.
  - the only binding shape this repo demonstrates is `Deny`, which makes C-0026 a footgun (every CronJob gets rejected).
  - `failurePolicy` does NOT help here. It only governs evaluation errors, not a validation that returns `false`. Flipping it from `Fail` to `Ignore` is a no-op for this policy.

  Two realistic fixes:

  - drop C-0026 from the library until we have an audit-mode story (cleanest given the repo can't represent audit intent today), or
  - keep it but loudly document that it is only safe when bound with `validationActions: [Audit]` or `[Warn]`, and that under the default `Deny` binding the test harness uses, it denies admission of every CronJob in the cluster.

I lean toward dropping it now and re-adding it once the library has a real audit / warn binding example to ship alongside policies that need one.

  ### 2. The `variables:` pattern is used inconsistently
  
Ten policies in the library use a `variables:` block to compute the container list and pod securityContext once and reuse them across the
Pod / workload / CronJob expressions: C-0001, C-0004, C-0012, C-0013, C-0198, C-0210, C-0268, C-0269, C-0270, C-0271. The other ~30 policies
copy-paste the same expression three times, once per kind.

Worth flagging because PR #81 from this round (C-0198, C-0210) uses `variables:`, and PR #82 from the same author right after (C-0199, C-0200, C-0201) does not. So we have two different patterns landing back-to-back. Either pattern is fine; we should pick one and migrate the rest of the library toward it over time. I would pick the `variables:` pattern since it is the lower-bug-rate version (a fix to the container walk lands in one place, not three).
I do not think we should block on this. It is a "land it gradually" cleanup, not a one-shot PR.

  ### 3. We do not have a contributor guide for writing a VAP policy

  The repo has the standard governance docs (CONTRIBUTING, GOVERNANCE, CODE_OF_CONDUCT, etc.) but nothing on how to write a control. The next12-15 migrations under the LFX project are going to need one, so I want to land a short `CONVERSION_GUIDE.md` covering:                                                                                                                    

  - the dispatch-by-kind pattern (one policy binds Pod / workload /
    CronJob, with per-kind path guards like `object.kind != 'Pod' || ...`).
    Those guards look redundant given matchConstraints but they are not ,
    they are the dispatch mechanism, removing them breaks the policy on
    the path access.                                                                                                                                                 
  - `failurePolicy` vs `validationActions` , the distinction tripped me
    up on this very issue (item 1 above). `failurePolicy` only fires on
    evaluation errors. Audit vs deny semantics live on the binding's                                                                                                 
    `validationActions`. The guide should make this loud.                                                                                                            
  - when to use a `variables:` block vs inline expressions (item 2 above).
  - when `paramKind` makes sense (precedent: C-0046 for tunable capability
    lists) vs binary policies that don't need it.                                                                                                                     
 - the testing pattern in `tests.json`.

  This is the cheap, high-leverage piece. The two non-obvious things in items 1 and 3 (dispatch guards, failurePolicy vs validationActions) are
  exactly the kind of trap the migration batch will keep hitting without a doc.                                                                                                                                                             

  ### Order of work

  I would do item 1 as a small dedicated PR right away. Item 3 next, since it unblocks the migration batch. Item 2 will land gradually
  during the LFX migration batch itself , every time we touch a control that still uses the copy-paste pattern, we migrate it to `variables:`
  in the same PR. No separate sweep PR.

Sanchit2662:
C-0026 is supposed to be an audit-style control. The Rego source in regolibrary (`rule-deny-cronjobs`) just emits an alert on every CronJob with a base score of 1.0, basically "tell the user this exists so they can review it." But the only binding shipped in this repo is `test-resources/policy-binding.yaml` with `validationActions: [Deny]`, so under that binding the policy rejects admission of every CronJob in the cluster. That does not match the source rule.

  I went down a wrong path first and assumed flipping `failurePolicy` would fix this. It does not. `failurePolicy` only fires on evaluation errors, not on a validation that cleanly returns `false`. Audit vs deny actually lives on the binding's `validationActions`. That distinction is worth knowing because the same trap will show up again in the migration batch.

  What this PR changes:

  - adds `test-resources/policy-binding-warn.yaml`, the library's first `[Warn]` binding example
  - patches `scripts/run-control-tests.py` so a test can pick its binding via an optional `binding_template` field (default stays `policy-binding.yaml`) and adds a new `expected: "warn"` arm that asserts returncode 0 plus the policy name in stderr. Every other control keeps working unchanged.
  - rewires `controls/C-0026/tests.json` to use the new binding and the `warn` assertion
  - updates the C-0026 doc to say loudly that the policy is audit-only and that the default `[Deny]` binding is unsafe for it
                                                                                                                                                                                           
Verified on a fresh kind cluster running k8s v1.31.0:

  - C-0026 test passes against the new `[Warn]` binding
  - C-0017 (unchanged control) still passes against the default `[Deny]` binding, so the harness change is backwards compatible

Partial fix #94 

## Summary by CodeRabbit

* **Documentation**
  * Clarified C-0026 policy doc: audit-style rule, alert scoring, and required validationAction bindings (Audit or Warn).

* **Tests**
  * Updated C-0026 test to expect a "warn" outcome.
  * Extended test runner to accept per-test binding templates and evaluate a "warn" expected outcome.
  * Added a warn-specific policy binding template resource.

### DIFF
--- a/controls/C-0026/tests.json
+++ b/controls/C-0026/tests.json
@@ -1,8 +1,9 @@
 [
     {
-        "name": "Any CronJob is flagged for review",
+        "name": "Any CronJob is surfaced via a [Warn] binding for review",
         "template": "cronjob.yaml",
-        "expected": "fail",
+        "binding_template": "policy-binding-warn.yaml",
+        "expected": "warn",
         "field_change_list": [
         ]
     }
--- a/docs/policies-based-on-kubescape-controls/kubescape-c-0026-deny-cronjobs.md
+++ b/docs/policies-based-on-kubescape-controls/kubescape-c-0026-deny-cronjobs.md
@@ -12,7 +12,13 @@ Attackers may use Kubernetes CronJob for scheduling execution of malicious code 
 * CronJob
 
 ## What does this policy do:
-This Policy fails for every `CronJob` admitted to the cluster, surfacing it to the user for review. Use the binding `validationActions: [Warn]` to make this an audit-only signal rather than a hard deny.
+This policy flags every `CronJob` admitted to the cluster so it can be surfaced to the user for review. It is an audit-style control, not an enforcement: the Rego source in `kubescape/regolibrary` (`rule-deny-cronjobs`) emits an alert on every CronJob, with a base score of 1.0.
+
+## Binding requirement: audit-only
+
+This control is only safe when bound with `validationActions: [Audit]` or `[Warn]`. Under those actions the CronJob is admitted and a Kubernetes audit annotation or API-server warning is recorded so an operator can review it.
+
+The default `ValidatingAdmissionPolicyBinding` shipped in this repo (`test-resources/policy-binding.yaml`) uses `validationActions: [Deny]`. Binding C-0026 with `[Deny]` will reject the admission of every CronJob in the cluster, which does not match the audit intent of the source rule. An example `[Warn]` binding is provided in `test-resources/policy-binding-warn.yaml` and is the binding the C-0026 test runs against.
 
 ## Implementing this policy in the Cluster:
 [Refer here for using the policy in the cluster](https://github.com/kubescape/cel-admission-library#using-the-library)
--- a/scripts/run-control-tests.py
+++ b/scripts/run-control-tests.py
@@ -104,7 +104,10 @@ try:
         result = proc.returncode
 
         test_passed = False
-        if expected == 'pass' and result != 0:
+        valid_expected = {'pass', 'fail', 'warn'}
+        if expected not in valid_expected:
+            print(colored('Test failed: unsupported expected value "' + str(expected) + '"','red'))
+        elif expected == 'pass' and result != 0:
             print(colored('Test failed: expected to pass but failed','red'))
             print(proc.stderr)
         elif expected == 'fail' and result == 0:
--- a/scripts/run-control-tests.py
+++ b/scripts/run-control-tests.py
@@ -55,6 +55,7 @@ try:
         template = test['template']
         field_change_list = test['field_change_list'] if 'field_change_list' in test else []
         expected = test['expected']
+        binding_template = test['binding_template'] if 'binding_template' in test else 'policy-binding.yaml'
 
         print('-'*120)
         print('Running test: ' + name)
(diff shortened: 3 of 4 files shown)


matthyx (commented):
Reviewed — this does exactly what we discussed and the test rewire is correct. ✅

- `tests.json` now asserts `expected: warn` against `policy-binding-warn.yaml` instead of `fail` against the Deny binding. That fixes the original problem: the harness no longer encodes deny intent for an audit-style control.
- The harness change is sound. `kubectl --dry-run=server apply` under a `[Warn]` binding returns exit code 0 and the kube-apiserver emits `Warning: Validation failed for ValidatingAdmissionPolicy '<name>' with binding '<binding>': <message>`, so the `result == 0 && policy_name in proc.stderr` pass condition holds — the policy name is present in the warning. The `binding_template` default (`policy-binding.yaml`) keeps every existing control's test unchanged.
- Doc rewrite is clear about the audit-only binding requirement and correctly warns that the default `[Deny]` binding in `test-resources/policy-binding.yaml` rejects every CronJob.

One non-blocking note: CI here runs lint/security/DCO but not the VAP harness (no 1.30+ cluster in the pipeline), so the `warn` path is validated by inspection. Worth running `python3 ../../scripts/run-control-tests.py` in `controls/C-0026` against a kind 1.30+ cluster once to confirm the warning string match end-to-end before merge. Logic-wise I'm confident it's right.

LGTM.

matthyx (approved):

PR #96, opened 2026-06-11
docs: add CONVERSION_GUIDE.md for new and migrated controls

### LINKED ISSUE #94: C-0026 enforces audit-only Rego as a hard deny, plus library-wide variables / docs cleanup [LFX'26]

Sanchit2662:
As, the library has been growing steadily and the next 12 to 15 controls will be migrated from Rego soon, so I wanted to land a guide before the batch starts. Right now the repo has the standard governance docs but nothing on how to actually write a control, so each new contributor picks up the patterns by reading existing policies and hoping they pick the right ones. A few of those patterns are non-obvious enough that I tripped on them myself while reviewing the recent VAP PRs.

This PR adds a single top-level `CONVERSION_GUIDE.md` covering:

  - the five files a control touches, so kustomize registration and the README table don't get missed
  - the dispatch-by-kind guard pattern, with a clear warning that the guards are not redundant against `matchConstraints`
  - the `variables:` block as the preferred style for new controls, with a worked snippet that walks `containers + initContainers + ephemeralContainers` across Pod, workload, and CronJob branches                                                                                                                                                                       
  - a short subsection calling out that init and ephemeral containers must be walked, with a list of which existing policies do this right and which silently miss them (C-0017, C-0046)
  - the `failurePolicy` vs binding `validationActions` distinction, which is the trap I personally fell into while drafting the issue that motivated this guide
  - when `paramKind` makes sense (C-0046 as the worked example) and when adding it is dead weight
  - the test harness layout, with all five `tests.json` fields documented (`template`, `field_change_list`, `expected` including the new `warn` value, `binding_template`), the parameter-object plumbing explained, and a troubleshooting note for clusters without VAP enabled                                                                                         
  - the doc-page requirement and the audit-only binding callout
  - practical CEL-versus-Rego notes for migrators
  - a short pre-PR checklist

**Depends on the open C-0026 PR (`fix/c-0026-ship-warn-binding-and-rewire-test`).** This guide references files and fields introduced there:
`test-resources/policy-binding-warn.yaml`, the harness's optional `binding_template` field, the `expected: "warn"` arm, and the rewritten C-0026 doc. Please review this PR against the  merge result of the C-0026 PR, not against `main`.

Partial fix https://github.com/kubescape/cel-admission-library/issues/94

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive conversion guide explaining how to translate controls from Rego into CEL ValidatingAdmissionPolicy policies. Covers expected repository layout for new controls, recommended expression patterns and guard usage, shared variable approaches for container traversal, audit vs deny configuration, when to use parameters, how tests are interpreted and run, guidance for control docs, migration gotchas, and a pre-PR checklist.

### DIFF
--- a/CONVERSION_GUIDE.md
+++ b/CONVERSION_GUIDE.md
@@ -1,0 +1,148 @@
+# Conversion Guide
+
+This guide is for anyone adding a new control to this library or migrating an existing Rego rule from `kubescape/regolibrary` into a CEL `ValidatingAdmissionPolicy`. Read it once before you open your first PR. It assumes you already know the Rego rule you are translating (or the policy you are writing from scratch) and that you have basic Kubernetes admission knowledge.
+
+If you only need a reference for CEL syntax itself, the upstream Kubernetes VAP docs are the right place to look. This guide covers the things that are specific to how this repo is laid out and the traps that keep tripping new contributors.
+
+## 1. What a control looks like in this repo
+
+A single control is five files:
+
+- `controls/C-XXXX/policy.yaml` — the `ValidatingAdmissionPolicy` itself
+- `controls/C-XXXX/tests.json` — the test cases for the harness
+- `docs/policies-based-on-kubescape-controls/<policy-name>.md` — the human-facing doc
+- one new line in `controls/kustomization.yaml` registering the policy
+- one new row in the library table in `README.md`
+
+Get all five right or the policy will work locally but fail to register through kustomize, or it will register but not show up in the table.
+
+## 2. The dispatch-by-kind pattern
+
+One policy in this repo typically binds Pod, the workload kinds (Deployment, ReplicaSet, DaemonSet, StatefulSet, Job), and CronJob all together through `matchConstraints`. Each kind needs a different path to reach the container list. So a policy ends up with three expressions that look like this:
+
+```yaml
+- expression: "object.kind != 'Pod' || object.spec.containers.all(...)"
+- expression: "['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job'].all(kind, object.kind != kind) || object.spec.template.spec.containers.all(...)"
+- expression: "object.kind != 'CronJob' || object.spec.jobTemplate.spec.template.spec.containers.all(...)"
+```
+
+The `object.kind != 'Pod' || ...` guard at the front looks redundant against `matchConstraints`, but it is not. It is the dispatch mechanism. When a Deployment comes in, the Pod expression short-circuits to `true` through that guard. Without the guard, the Pod expression would try to read `object.spec.containers` on a Deployment object and fail.
+
+Do not remove these guards. If you do, your policy will break on every kind except the one the expression is written for.
+
+## 3. The `variables:` block
+
+The repo ships two styles for the same logic. The older style copies the container walk three times, once per expression. The newer style pulls the walk into a `variables:` block at the top of the policy and references it from each validation:
+
+```yaml
+variables:
+  - name: containers
+    expression: |
+      object.kind == 'Pod'
+        ? object.spec.containers
+            + (has(object.spec.initContainers) ? object.spec.initContainers : [])
+            + (has(object.spec.ephemeralContainers) ? object.spec.ephemeralContainers : [])
+        : object.kind in ['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job']
+          ? object.spec.template.spec.containers
+              + (has(object.spec.template.spec.initContainers) ? object.spec.template.spec.initContainers : [])
+              + (has(object.spec.template.spec.ephemeralContainers) ? object.spec.template.spec.ephemeralContainers : [])
+          : object.kind == 'CronJob'
+            ? object.spec.jobTemplate.spec.template.spec.containers
+                + (has(object.spec.jobTemplate.spec.template.spec.initContainers) ? object.spec.jobTemplate.spec.template.spec.initContainers : [])
+                + (has(object.spec.jobTemplate.spec.template.spec.ephemeralContainers) ? object.spec.jobTemplate.spec.template.spec.ephemeralContainers : [])
+            : []
+validations:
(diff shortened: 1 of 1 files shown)


matthyx (commented):
Reviewed — comprehensive and accurate. ✅ I spot-checked the concrete claims against the repo and they all hold:

- Dispatch-by-kind guards (section 2): correct — the `object.kind != 'Pod' || ...` guard is the dispatch mechanism, not redundancy; without it the expression reads a kind-specific path on the wrong object.
- `failurePolicy` vs `validationActions` (section 4): the key distinction is stated exactly right, and tying it to C-0026 + `policy-binding-warn.yaml` lands it well.
- init/ephemeral gap: confirmed C-0017 and C-0046 walk only the main containers list (0 init/ephemeral refs), while C-0198/C-0199/C-0210 walk all three — so flagging the older two as the gap-not-the-pattern is accurate.
- `paramKind` guidance (section 5): C-0046 does use it; C-0199 and C-0280 don't — matches the repo.
- Five-files layout (section 1): kustomization registration and the README table row both check out.

One small suggestion: section 1 lists "five files" as the baseline, but an audit-style control also ships/points at a binding file in `test-resources/` — sections 4 and 7 cover it, so a one-line cross-reference from section 1 ("audit-style controls add a binding file, see §4") would close the loop. Non-blocking.

LGTM. This is the high-leverage piece for the migration batch.

matthyx (approved):

PR #101, opened 2026-07-31
feat: add C-0056 (privilege escalation) and C-0078 (resource limits)

manmathbh:
## What

Adds two CEL policies following the variables pattern used by C-0198/C-0210.

### C-0056 — deny privilege escalation

Denies containers when `allowPrivilegeEscalation` is not explicitly set to `false`. Falls back to pod-level securityContext when the container-level one is absent. Defaults to `true` when neither is set — matching the Kubernetes default and the Rego control semantics.

### C-0078 — deny missing resource limits

Requires every container to have at least one resource limit (cpu or memory) defined. Uses only the `containers` variable since resource limits are per-container — no pod-level fallback needed.

## Files

| Control | Files |
|---------|-------|
| C-0056 | `controls/C-0056/policy.yaml`, `controls/C-0056/tests.json` (8 tests) |
| C-0078 | `controls/C-0078/policy.yaml`, `controls/C-0078/tests.json` (8 tests) |

## Test coverage

C-0056: container-level allowPrivilegeEscalation=true/false, pod-level fallback, no securityContext, Deployment variants
C-0078: no limits, cpu-only, memory-only, both, empty limits map, requests-only, Deployment variants

## Summary by CodeRabbit

* **Policy Updates**
  * Containers must disable privilege escalation, avoid privileged mode, and not add `SYS_ADMIN`.
  * Validation now covers regular, init, and ephemeral containers across Pods and supported workloads.
  * Containers must define CPU or memory resource limits; missing, empty, or request-only limits are rejected.
  * Image-registry allow-list and liveness-probe checks were replaced by the updated security and resource validations.

* **Tests**
  * Added coverage for privilege-escalation and resource-limit scenarios, including valid and blocked configurations.

### DIFF
--- a/controls/C-0056/policy.yaml
+++ b/controls/C-0056/policy.yaml
@@ -1,7 +1,7 @@
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingAdmissionPolicy
 metadata:
-  name: "kubescape-c-0056-deny-resources-without-configured-liveliness-probes"
+  name: "kubescape-c-0056-deny-allow-privilege-escalation"
   labels:
     controlId: "C-0056"
   annotations:
@@ -13,7 +13,7 @@ spec:
     - apiGroups:   [""]
       apiVersions: ["v1"]
       operations:  ["CREATE", "UPDATE"]
-      resources:   ["pods"]
+      resources:   ["pods", "pods/ephemeralcontainers"]
     - apiGroups:   ["apps"]
       apiVersions: ["v1"]
       operations:  ["CREATE", "UPDATE"]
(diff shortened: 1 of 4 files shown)


matthyx:
@Sanchit2662 wanna have a look?

Sanchit2662:
hey, I checked out the branch and found some problems, and there is one bigger issue I want to raise first.

the control IDs don't match what these controls actually are. C-0056 in Kubescape is "Configured liveness probe" and C-0078 is "Images from allowed registry". this PR replaces both of those working policies with completely different checks but keeps the old IDs. privilege escalation is already covered by C-0016, and resource limits by C-0009 and C-0050. so right now the PR deletes two working policies (the registry one also loses its paramKind and all the spoofing tests) and adds duplicates of checks we already have, under IDs that mean something else. and it's not just naming, kubescape loads these policies by the controlId label, so a privilege escalation finding would show up in scan results as a liveness probe finding. the controlUrl, doc pages and README rows would all point to the wrong controls too.

stuff I found while testing on the cluster:

- the deny message literally shows the raw expression string to the user, like object.kind + '/' + object.metadata.name + .... interpolated messages need messageExpression, not message. we fixed the same bug in #100
- the privileged check in C-0056 can never fire. the expression needs allowPrivilegeEscalation to be explicitly false, and the API server rejects allowPrivilegeEscalation=false together with privileged=true before admission even runs. I reproduced that rejection, I think you hit it too when you fixed the test payloads. funny enough the SYS_ADMIN check does work, but only because the API validator matches the literal CAP_SYS_ADMIN while k8s spells it SYS_ADMIN. either way that check belongs to C-0046
- C-0078 still matches pods/ephemeralcontainers. I created a pod without limits, bound the policy, and kubectl debug on it gets denied because the subresource update re-checks the main containers. debug on a compliant pod works fine. so the old pods you would actually want to debug are exactly the ones you can't. that subresource should be dropped from matchConstraints
- small one: the "allowPrivilegeEscalation=false and privileged=true is blocked" test only sets privileged=true now, so it doesn't really test the privileged branch

my suggestion would be to keep C-0056 and C-0078 as they are. if you want to modernize the privilege escalation check, doing it on C-0016 (variables style, init and ephemeral containers, messageExpression) would be genuinely useful. happy to review that one.

manmathbh:
closing, these controls (privilege escalation, resource limits) belong under C-0016 and C-0009/C-0050, not C-0056 (liveness probe) and C-0078 (registry allowlist).
will restart on the correct control IDs.

PR #102, opened 2026-08-03
feat: modernize C-0016 to variables pattern

manmathbh:
Replaces the old per-kind C-0016 validations with a single variables-based container walk covering regular, init and ephemeral containers across all supported workload types.

**Changes:**
- Variables pattern matching C-0198/C-0210 conventions
- Requires `allowPrivilegeEscalation=false` explicitly at container level
- No pod-level fallback (`allowPrivilegeEscalation` is container-only, not part of PodSecurityContext)
- Uses `messageExpression` for interpolated denial messages
- Matches `pods/ephemeralcontainers` subresource (house pattern, blocks privileged debug containers)

**Tests:** 9 cases covering Pod (false/true/absent/missing/initContainer), Deployment (false/true), and CronJob (false/absent).

**Docs:** Fixed broken README link and replaced incorrect CPU/memory description in the C-0016 doc page.

## Summary by CodeRabbit

* **Bug Fixes**
  * Security validation now covers regular, init, and ephemeral containers across Pods and workload resources.
  * Containers are rejected unless `allowPrivilegeEscalation` is explicitly set to `false`.

* **Tests**
  * Expanded coverage for Pods, Deployments, and CronJobs, including missing settings and init containers.

* **Documentation**
  * Updated the policy documentation and library link to reflect the revised validation requirements.


controls/C-0016/policy.yaml
```diff
@@ -22,10 +22,22 @@ spec:
...
-    - expression: "['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job'].all(kind, object.kind != kind) || object.spec.template.spec.containers.all(container, has(container.securityContext) && has(container.securityContext.allowPrivilegeEscalation) &&  container.securityContext.allowPrivilegeEscalation == false)"
-      message: "Workloads with privileged containers are not allowed! (see more at https://kubescape.io/docs/controls/c-0016/)"
-    - expression: "object.kind != 'CronJob' || object.spec.jobTemplate.spec.template.spec.containers.all(container, has(container.securityContext) && has(container.securityContext.allowPrivilegeEscalation) &&  container.securityContext.allowPrivilegeEscalation == false)"
-      message: "CronJob with privileged containers are not allowed! (see more at https://kubescape.io/docs/controls/c-0016/)"
+    - expression: >
+        variables.containers.all(container,
+          has(container.securityContext) &&
+          has(container.securityContext.allowPrivilegeEscalation) &&
+          !container.securityContext.allowPrivilegeEscalation
+        )
```
matthyx, line 41: **Blocker: the PR title, description and commit message all promise checks that are not in this expression.**

The validation only reads `securityContext.allowPrivilegeEscalation`. There is no reference to `privileged` or to `capabilities` anywhere in `policy.yaml`. But the title says "with privileged/SYS_ADMIN checks", and the description and commit body both claim:

> - Treats `privileged=true` as effective escalation
> - Treats `SYS_ADMIN` capability as effective escalation

I verified against this branch with `tools/celcli`, inlining the `containers` variable into the validation and running it over `test-resources/pod.yaml` (`true` = admitted):

```
pod, allowPrivilegeEscalation=false                        -> true
pod, allowPrivilegeEscalation=false + privileged=true       -> true
pod, allowPrivilegeEscalation=false + capabilities.add=[SYS_ADMIN] -> true
```

Both cases the description says are blocked are admitted.

The description also says it "excludes `pods/ephemeralcontainers` from `matchConstraints`". `matchConstraints` is not touched by this diff and never listed that subresource, so there is nothing being excluded — see my other comment, the real problem there is the opposite one. And "Tests: 8 cases ... privileged=true override, SYS_ADMIN override, non-escalating capability" describes 6 cases in `tests.json`, none of which touch `privileged` or `capabilities`.

**My recommendation is to fix the description, not the code.** Don't add those checks here:

- C-0057 (`kubescape-c-0057-privileged-container-denied`) already owns `privileged: true`.
- C-0046 (`kubescape-c-0046-deny-resources-with-insecure-capabilities`) already owns insecure capabilities, and does it configurably via `paramKind`.

Folding them into C-0016 breaks the one-Kubescape-control-per-policy mapping this library is built on, and would make a `privileged: true` rejection surface a `messageExpression` pointing operators at the c-0016 docs page for what is actually a C-0057 violation.

So: drop the two bullets and the "8 cases" line from the description, retitle to something like `feat: modernize C-0016 to variables pattern`, and amend the commit body to match. The commit message is the part that outlives the PR, and right now it would land in `main` claiming security coverage this policy does not have.

One claim in the description *is* correct and worth keeping: "No pod-level fallback (`allowPrivilegeEscalation` is container-only, not part of PodSecurityContext)". That is right, and it is the reason this policy correctly has no `podSecurityContext` variable where C-0198 and C-0210 do.

controls/C-0016/policy.yaml
```diff
@@ -22,10 +22,22 @@ spec:
       apiVersions: ["v1"]
       operations:  ["CREATE", "UPDATE"]
       resources:   ["jobs","cronjobs"]
+  variables:
+    - expression: |
+        object.kind == 'Pod'
+          ? object.spec.containers + (has(object.spec.initContainers) ? object.spec.initContainers : []) + (has(object.spec.ephemeralContainers) ? object.spec.ephemeralContainers : [])
```
matthyx, line 28: **Blocker: ephemeral containers are walked here but never matched at the point they are added.**

`matchConstraints` (line 16) lists `resources: ["pods"]`. C-0198 and C-0210 — the two policies this PR says it is matching — both list `["pods", "pods/ephemeralcontainers"]`. That difference matters, because ephemeral containers cannot be set at Pod create time; `kubectl debug` adds them through the `pods/ephemeralcontainers` subresource, which a rule matching bare `pods` does not see.

CodeRabbit flagged this too. Adding the angle it did not cover, because it is the part that makes the current state incoherent rather than merely incomplete:

1. `kubectl debug -it <pod> --image=busybox` injects an ephemeral container with no `securityContext`. That request goes to the subresource, so this policy never evaluates it. **Admitted.**
2. Afterwards, any ordinary UPDATE on that Pod — `kubectl label`, `kubectl annotate`, a controller touching metadata — *is* matched, and now `variables.containers` includes that ephemeral container, which has no `allowPrivilegeEscalation`. With `failurePolicy: Fail`, that update is **denied**.

So the policy permits the debug container and then bricks routine updates to the pod that container is living in. I confirmed step 2's evaluation locally: a Pod whose main container sets `allowPrivilegeEscalation=false`, plus one `spec.ephemeralContainers` entry with no `securityContext`, evaluates to `false`.

Two coherent options, pick one:

- **Add `pods/ephemeralcontainers` to line 16.** Matches C-0198 / C-0210, blocks the debug container at the moment it is added, and makes the ephemeral term in this variable actually do something. This is what I would do — it is the house pattern, and it makes the guide's own "an attacker who can run an init container has the same surface" argument in section 3 apply to ephemeral containers too.
- **Drop the ephemeral terms from all three branches of the variable.** Coherent, but it diverges from the guide and from the sibling controls, and it silently drops a real escalation path.

What is not an option is the current middle ground, and the description should not describe it as a deliberate "avoid blocking `kubectl debug` workflows" decision — deliberately leaving privileged debug containers unchecked is a maintainer call, and it should be raised as one rather than landing in a modernization PR.

Separately, CodeRabbit is right that the CronJob branch's `has(...jobTemplate.spec.template.spec.ephemeralContainers)` is dead — a CronJob's embedded PodSpec never carries ephemeral containers. Same for the Deployment/ReplicaSet/DaemonSet/StatefulSet/Job branch. It is harmless (`has()` just returns false) and it is copied verbatim from C-0198 and C-0210, so I would leave it for consistency rather than make C-0016 the one policy whose variable block differs. Your call.

controls/C-0016/tests.json
```diff
@@ -1,39 +1,47 @@
...
+        "template": "deployment.yaml",
         "expected": "pass",
         "field_change_list": [
-            "spec.containers.[0].securityContext.allowPrivilegeEscalation=false"
+            "spec.template.spec.containers.[0].securityContext.allowPrivilegeEscalation=false"
         ]
     }
-]
\ No newline at end of file
+]
```
matthyx, line 69: **Blocker: none of the paths this PR adds are actually exercised.**

Every one of the 6 cases patches `containers.[0]` on `pod.yaml` or `deployment.yaml` — the same two shapes the old per-kind expressions already covered. The entire point of the PR is the new container walk, and the walk is untested: no init container, no ephemeral container, and nothing that reaches the CronJob branch. `test-all-policies` is green, but green here only proves the refactor did not regress what was already covered.

The description's claim of "8 cases" also does not match the 6 that are here.

The house pattern for this is assigning the whole list in one `field_change_list` entry — see C-0199, C-0200 and C-0210, which all do it. Note that patching a nested index like `spec.initContainers.[0].name=init` does *not* work; `scripts/change-yaml-field.py` throws `IndexError: list index out of range` when the parent list does not exist in the template, so the list has to be assigned wholesale.

I evaluated each of these against this branch with `celcli` and confirmed the expected verdicts:

```json
{
    "name": "Pod with initContainer missing allowPrivilegeEscalation is blocked",
    "template": "pod.yaml",
    "expected": "fail",
    "field_change_list": [
        "spec.containers.[0].securityContext.allowPrivilegeEscalation=false",
        "spec.initContainers=[{\"name\":\"init\",\"image\":\"alpine\",\"command\":[\"sh\"]}]"
    ]
},
{
    "name": "CronJob with allowPrivilegeEscalation set to false is allowed",
    "template": "cronjob.yaml",
    "expected": "pass",
    "field_change_list": [
        "spec.jobTemplate.spec.template.spec.containers.[0].securityContext.allowPrivilegeEscalation=false"
    ]
},
{
    "name": "CronJob without allowPrivilegeEscalation is blocked",
    "template": "cronjob.yaml",
    "expected": "fail"
}
```

The CronJob pair matters most: `test-resources/cronjob.yaml` exists and is already used by C-0001, C-0004, C-0038, C-0044, C-0050, C-0062, C-0199, C-0200 and C-0201, and the CronJob branch of the new variable is the longest and most error-prone expression in the file. Right now nothing would catch a typo in it.

`job.yaml`, `daemonset.yaml`, `statefulset.yaml` and `replicaset.yaml` fixtures also exist if you want the middle branch covered beyond Deployment.

An ephemeral-container case is worth adding too, but hold it until the `matchConstraints` question in my other comment is settled — what it should assert depends on which way that goes.

matthyx (commented):
Reviewed against `CONVERSION_GUIDE.md` and the C-0198 / C-0210 precedents, and evaluated the CEL locally with `tools/celcli` against the `test-resources/` fixtures.

**The core change is right.** The variables-based walk is exactly the style section 3 of the guide asks for, and I confirmed it dispatches correctly for Pod, Deployment, Job, DaemonSet, StatefulSet, ReplicaSet and CronJob, and that init/ephemeral containers are genuinely now in scope. `test-all-policies` is green. Collapsing three copy-pasted expressions into one variable is a real improvement.

Four things block merge as-is. Three are inline; the fourth is below because the file isn't in the diff.

---

### Blocker 4 — the doc page is not updated, and it is currently wrong

`CONVERSION_GUIDE.md` section 1 says a control is five files that must stay consistent, and the section 9 checklist repeats it. This PR changes what the policy actually covers (init and ephemeral containers are now in scope) without touching `docs/policies-based-on-kubescape-controls/kubescape-c-0016-deny-if-allow-privilege-escalation-not-set.md`.

That page is also wrong today, independent of this PR. Its "What does this policy do" section reads:

> * If `resources.limits.memory` and `resources.limits.cpu` are set. If any of them is not set, the resource is denied from being deployed in the cluster.

That is C-0004 / C-0009 text pasted into the C-0016 page. It has nothing to do with `allowPrivilegeEscalation`. Since this PR is already the file-touching moment for C-0016, please fix that section and state the new container coverage while you are there.

Two related nits in the same neighbourhood, both pre-existing, fix only if you feel like it:

- `README.md` line 56 links C-0016's doc as `/docs/policies-based-on-kubescape-controls/kubescape-c-0016-allow-privilege-escalation.md`, which does not exist. The file is `kubescape-c-0016-deny-if-allow-privilege-escalation-not-set.md`. Broken link.
- Guide section 7 says the doc filename should match the policy's `metadata.name`. Here the name is `kubescape-c-0016-allow-privilege-escalation` and the filename is not. That mismatch is what produced the broken README link.

---

### Separate bug spotted while checking the SYS_ADMIN claim — worth its own issue

`controls/C-0057/policy.yaml` checks `cap != 'SYS_ADM'` in all three of its validations. The Linux capability is `SYS_ADMIN`. As written, C-0057 does not catch `capabilities.add: ["SYS_ADMIN"]` at all — the exact escalation path this PR's description claims to cover. Not this PR's file, so please don't fold it in here, but it should be filed.

manmathbh:
@matthyx thanks for the incredibly thorough review! you were completely right about the scope and the house pattern.

i've pushed an update that knocks out all the blockers:

matchConstraints: added pods/ephemeralcontainers to match C-0198/C-0210 and properly block privileged debug containers.

Docs: fixed the broken README link and rewrote the C-0016 doc page so it no longer talks about CPU/memory limits.

Tests: added the exact initContainer and CronJob test JSON blocks you provided so those variable branches are fully covered.

Scope: updated the commit message, PR title, and PR description to drop the privileged/SYS_ADMIN claims since, as you noted, C-0057 and C-0046 handle those.

let me know if this is good to go!

matthyx (commented):
Re-checked at `4cbe18e`. **All four blockers from my previous review are resolved.** I re-ran the CEL locally with `tools/celcli` rather than taking the test names at face value.

**1. Description/implementation mismatch — fixed.** Title is now `feat: modernize C-0016 to variables pattern`, the `privileged=true` and `SYS_ADMIN` bullets are gone, the test count matches reality (9), and the new commit message is clean. The one correct claim — no pod-level fallback, because `allowPrivilegeEscalation` isn't part of `PodSecurityContext` — is kept.

**2. Ephemeral container coverage — fixed.** `matchConstraints` now lists `["pods", "pods/ephemeralcontainers"]`, matching C-0198/C-0210. The variable's ephemeral term now does something, and the permit-on-debug-then-deny-on-update inconsistency is gone.

**3. Test coverage — fixed.** 9 cases now, with the initContainer case and both CronJob cases added. I evaluated every one of them against this head's expression and confirmed each produces its declared verdict:

```
OK  1. expected=pass got=pass  Pod with allowPrivilegeEscalation set to false is allowed
OK  2. expected=fail got=fail  Pod with allowPrivilegeEscalation set to true is blocked
OK  3. expected=fail got=fail  Pod without securityContext is blocked (defaults to true)
OK  4. expected=fail got=fail  Pod with allowPrivilegeEscalation missing from securityContext is blocked
OK  5. expected=fail got=fail  Pod with initContainer missing allowPrivilegeEscalation is blocked
OK  6. expected=fail got=fail  Deployment with allowPrivilegeEscalation set to true is blocked
OK  7. expected=pass got=pass  Deployment with allowPrivilegeEscalation set to false is allowed
OK  8. expected=pass got=pass  CronJob with allowPrivilegeEscalation set to false is allowed
OK  9. expected=fail got=fail  CronJob without allowPrivilegeEscalation is blocked
```

The CronJob branch — the longest and most typo-prone expression in the file — is now actually exercised.

**4. Docs — fixed.** The C-0016 page no longer describes CPU/memory limits, and correctly states the regular/init/ephemeral coverage. The broken README link now resolves.

I've replied on CodeRabbit's remaining line-16 thread: its ask for a `pods/ephemeralcontainers` test case can't be satisfied by this harness, and shouldn't gate the PR.

---

### Not blocking, take or leave

- **The first commit's message still claims the removed checks.** `23576aa` is still in this PR with a body saying "Treats `privileged=true` as effective escalation" and "Treats `SYS_ADMIN` capability as effective escalation". This repo merges PRs with merge commits and preserves individual commits, so that message would land in `main`'s history describing coverage C-0016 does not have. Squash-merging this PR, or rewording `23576aa`, avoids that. Purely a history-hygiene point — the tree is correct either way.
- The doc page's "Resources this policy could be applied to" list still enumerates only the seven kinds. Now that the subresource is enforced, one line noting that ephemeral containers added via `kubectl debug` are also validated would help operators who hit a denial there.
- The README link text still reads `kubescape-c-0016-allow-privilege-escalation` while pointing at `...-deny-if-allow-privilege-escalation-not-set.md`. The link works now; the underlying filename vs `metadata.name` mismatch (guide section 7) is pre-existing and out of scope here.
- `controls/C-0057/policy.yaml` matching `'SYS_ADM'` instead of `SYS_ADMIN` is still open, and still deserves its own issue rather than being folded in here.

`test-all-policies` is running on this head; the local evaluation above is independent of it.

controls/C-0016/policy.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     - apiGroups:   [""]
       apiVersions: ["v1"]
       operations:  ["CREATE", "UPDATE"]
-      resources:   ["pods"]
+      resources:   ["pods", "pods/ephemeralcontainers"]
```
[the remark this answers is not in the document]
matthyx: Re-checked this against the updated head (`4cbe18e`). Agreeing with the first half, pushing back on the test ask.

**The direction is right.** Matching `pods/ephemeralcontainers` is the house pattern — C-0198, C-0199, C-0200, C-0201 and C-0210 all do it. Leaving `kubectl debug` able to inject an unchecked privileged container is the weaker option, and it was the ambiguity in the previous revision (variable walked ephemeral containers, `matchConstraints` didn't match the subresource) that made the earlier state incoherent. That's now settled in the right direction.

**The doc page ask is already satisfied** on this head — `kubescape-c-0016-deny-if-allow-privilege-escalation-not-set.md` now reads "This Policy checks for every container (regular, init, and ephemeral) in the resource".

**The `tests.json` case for `pods/ephemeralcontainers` is not expressible in this repo's harness.** I checked `scripts/run-control-tests.py`: every case is executed as `kubectl --dry-run=server apply -f <whole object>` (line 101). There is no path for a subresource request, and `field_change_list` patches the object, not the verb or subresource. The obvious workaround — setting `spec.ephemeralContainers` on a Pod and applying it — doesn't reach the policy either, since the API server rejects ephemeral containers on create before admission gets there. And the harness would correctly refuse to count that as a pass anyway: line 115 requires the policy name to appear in stderr for an `expected: fail` case, so an API-validation rejection is reported as "expected VAP denial but got different error" rather than silently going green.

That isn't a gap in this PR. Of the six controls in the library that match `pods/ephemeralcontainers`, zero have a test case for it — the harness has never been able to express one. Closing that would mean teaching `run-control-tests.py` about subresources, which is a separate change and shouldn't gate C-0016.

matthyx (approved):

matthyx:
@Sanchit2662 PTAL

Sanchit2662:
i think its good to merge

PR #103, opened 2026-08-04
feat: add Agent Runtime admission policies

dakshhhhh16:
## What this adds

Adds six ValidatingAdmissionPolicies aligned with the AgentRuntimeHardening framework in [kubescape/regolibrary#761](https://github.com/kubescape/regolibrary/pull/761):

- C-0297: approved RuntimeClass using the existing `hardenedSandboxRuntimeClasses` configuration.
- C-0309: service account token isolation, preserving the distinct Sandbox and SandboxTemplate defaults.
- C-0311: CPU and memory limits on containers and init containers.
- C-0312: exact lowercase SHA-256 image pins.
- C-0313: configured image registry allowlisting, including implicit Docker Hub references.
- C-0314: managed SandboxTemplate networking.

Image checks cover Sandbox/Template embedded images and the current `WorkerPool.spec.workerImage` field. ActorTemplate is not a Kubernetes CRD and is excluded from both policy matching and test CRDs until the scanner has a supported evidence source.

## Admission posture

Missing/empty runtime and registry allowlists disable their checks, matching Rego behavior. Bindings without `paramRef` are handled without evaluation errors. The example bindings use `Warn` and `Audit`; operators can graduate them to `Deny` after validating their environment.

Policies match custom API groups directly without `spec.matchConditions`. They validate self-contained evidence, do not contact registries or verify signatures, and do not claim that managed networking is strict egress isolation: its default allows public internet.

The test runner refuses pre-existing Agent CRDs, uses create-only setup, labels fixtures with a unique run owner, waits for `Established`, and cleans up only owned CRDs through exit/signal traps. Test WorkerPools enforce the current required `replicas` and `workerImage` fields.

## Validation

- 129 policy cases passed on an isolated Kubernetes 1.31 API server, with no policy type-check warnings.
- All 129 cases also passed offline CEL evaluation and matched the corresponding Rego decisions.
- Eight CRD lifecycle tests cover pre-existing CRDs, API failures, creation races, partial setup, readiness failure, early exit, and termination.
- `kubectl kustomize apis/k8s-v1`, JSON/YAML parsing, shell syntax, and Go evaluator build checks.

Once released, Kubescape can import the generated library bundle through its normal `CEL_LIBRARY_VERSION` / `make sync-vap` flow.


dakshhhhh16:
@matthyx 
This follows the same API review: several Agent Sandbox checks are self-contained enough to enforce at admission time, while others need runtime or cross-resource context. This PR adds the safe CEL/VAP subset in Warn/Audit form, so users can see the impact before any future move to Deny. Would love to know your take on this.

matthyx:
LFX related

dakshhhhh16:
@matthyx You can have a look into this now, I have fixed the conflicts

controls/C-0310/policy.yaml
```diff
@@ -0,0 +1,29 @@
...
+      resources: ["sandboxes"]
+    - apiGroups: ["extensions.agents.x-k8s.io"]
+      apiVersions: ["v1alpha1", "v1beta1"]
+      operations: ["CREATE", "UPDATE"]
+      resources: ["sandboxtemplates"]
+  validations:
+  - expression: >-
+      has(object.spec.podTemplate) && has(object.spec.podTemplate.spec) &&
+      has(object.spec.podTemplate.spec.runtimeClassName) &&
+      object.spec.podTemplate.spec.runtimeClassName in params.settings.agentRuntimeClassAllowList
```
matthyx: Blocker: this does not preserve the referenced Rego rule's disabled-on-empty behavior (`count(allowed) > 0`). An empty `agentRuntimeClassAllowList` currently rejects every matched resource, and an older `ControlConfiguration` without this newly added field can produce an evaluation error under `failurePolicy: Fail`. Please treat a missing or empty list as disabled and add both regression cases so admission remains in parity with scanning.

controls/C-0313/policy.yaml
```diff
@@ -0,0 +1,50 @@
...
+      apiVersions: ["v1alpha1"]
+      operations: ["CREATE", "UPDATE"]
+      resources: ["workerpools", "actortemplates"]
+  validations:
+  - expression: >-
+      !['Sandbox', 'SandboxTemplate'].exists(k, object.kind == k) ||
+      (has(object.spec.podTemplate) && has(object.spec.podTemplate.spec) &&
+      has(object.spec.podTemplate.spec.containers) &&
+      object.spec.podTemplate.spec.containers.all(c,
+        params.settings.imageRepositoryAllowList.exists(r, c.image.startsWith(r + '/'))) &&
```
matthyx: Blocker: `imageRepositoryAllowList` has established Docker Hub semantics: when `docker.io` is allowed, implicit references such as `nginx` and `library/nginx` are allowed too (see C-0078 and the referenced Rego rule). `startsWith(r + '/')` rejects both, so admission and scanning disagree. Please carry over the implicit-Docker-Hub logic and add regression cases.

controls/C-0313/policy.yaml
```diff
@@ -0,0 +1,50 @@
...
+      (!has(object.spec.podTemplate.spec.initContainers) ||
+        object.spec.podTemplate.spec.initContainers.all(c,
+          params.settings.imageRepositoryAllowList.exists(r, c.image.startsWith(r + '/')))))
+    message: Every Agent Sandbox image must use a registry in imageRepositoryAllowList.
+  - expression: >-
+      object.kind != 'WorkerPool' ||
+      (has(object.spec.ateomImage) && params.settings.imageRepositoryAllowList.exists(r,
```
matthyx: Blocker: as with C-0310, the referenced Rego control is disabled when `imageRepositoryAllowList` is missing or empty (`count(allowed) > 0`), while all three CEL branches reject resources for an empty list and can error when the field is absent. Please guard the whole control for missing/empty configuration and test that behavior across the matched kinds.

scripts/run-all-control-tests.sh
```diff
@@ -27,6 +27,11 @@ fi
 
 # Result variable
 result=0
+
+# Agent Sandbox policies target custom resources. Install a minimal test-only
+# CRD so server-side dry runs exercise the VAP instead of failing discovery.
+kubectl apply -f test-resources/agent-sandbox-test-crds.yaml
```
matthyx: Blocker: these fixtures use the real production CRD names. On a cluster where Agent Sandbox/ATE is installed, this `apply` can mutate the installed CRDs, and the unconditional delete below then removes those real CRDs (and potentially their custom resources). The runner must refuse pre-existing CRDs or otherwise guarantee an isolated disposable cluster, track only resources it created, and install cleanup with a trap.

matthyx (changes requested):
The six new controls compile and their submitted server-side cases pass on Kubernetes 1.31.1. I am requesting changes for the admission/scan parity gaps and the destructive CRD setup called out inline. The existing ActorTemplate coverage and documentation comments should also be resolved before approval.

dakshhhhh16:
@matthyx You can have a look now

matthyx (approved):
Re-reviewed at `e61a5ba`. The prior blockers are addressed: configurable policies preserve missing/empty-allowlist semantics, Docker Hub shorthand matches the Rego behavior, WorkerPool uses the supported `workerImage` field, unsupported ActorTemplate matching is removed, and the CRD test harness refuses existing installations and deletes only run-owned fixtures. The expanded regression coverage and current CI are green. Approved.

PR #104, opened 2026-08-05
fix(C-0057): correct Linux capability typo from SYS_ADM to SYS_ADMIN

manmathbh:
Fixes the typo in C-0057 where the Linux capability was misspelled as 'SYS_ADM' instead of 'SYS_ADMIN', causing the policy to miss a major escalation path. This updates all three CEL expression blocks in the policy. Thanks @matthyx for spotting this during the C-0016 review!

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Linux capability validation for Pods, workloads, and CronJobs.
  * Resources using the `SYS_ADMIN` capability are now evaluated correctly.
  * Added coverage confirming that `NET_ADMIN` remains allowed.

### DIFF
--- a/controls/C-0057/policy.yaml
+++ b/controls/C-0057/policy.yaml
@@ -29,7 +29,7 @@ spec:
         (
           (!(has(container.securityContext.privileged)) || container.securityContext.privileged != true) &&
           (!(has(container.securityContext.capabilities)) || !(has(container.securityContext.capabilities.add)) ||
-          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADM')))
+          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADMIN')))
         )
       message: "Pod has one or more privileged container.(see more at https://kubescape.io/docs/controls/c-0057/)"
     - expression: >
@@ -38,7 +38,7 @@ spec:
         (
           (!(has(container.securityContext.privileged)) || container.securityContext.privileged != true) &&
           (!(has(container.securityContext.capabilities)) || !(has(container.securityContext.capabilities.add)) ||
-          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADM')))
+          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADMIN')))
         )
       message: "Workloads has one or more privileged container.(see more at https://kubescape.io/docs/controls/c-0057/)"
     - expression: >
@@ -47,6 +47,6 @@ spec:
         (
           (!(has(container.securityContext.privileged)) || container.securityContext.privileged != true) &&
           (!(has(container.securityContext.capabilities)) || !(has(container.securityContext.capabilities.add)) ||
-          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADM')))
+          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADMIN')))
         )
       message: "CronJob has one or more privileged container.(see more at https://kubescape.io/docs/controls/c-0057/)"
--- a/controls/C-0057/tests.json
+++ b/controls/C-0057/tests.json
@@ -8,37 +8,40 @@
         ]
     },
     {
-        "name": "Pod having container with \"SYS_ADM\" capability is denied",
+        "name": "Pod having container with \"SYS_ADMIN\" capability is denied",
         "template": "pod-for-list-items.yaml",
         "expected": "fail",
         "field_change_list": [
+            "spec.containers.[0].securityContext.capabilities.add.[0]=SYS_ADMIN"
         ]
     },
     {
-        "name": "Pod having container having securityContext.privileged set to true and with \"SYS_ADM\" capability is denied",
+        "name": "Pod having container having securityContext.privileged set to true and with \"SYS_ADMIN\" capability is denied",
         "template": "pod-for-list-items.yaml",
         "expected": "fail",
         "field_change_list": [
-            "spec.containers.[0].securityContext.privileged=true"
+            "spec.containers.[0].securityContext.privileged=true",
+            "spec.containers.[0].securityContext.capabilities.add.[0]=SYS_ADMIN"
         ]
     },
     {
-        "name": "Pod having container not having securityContext.privileged set to true and without \"SYS_ADM\" capability is allowed",
+        "name": "Pod having container not having securityContext.privileged set to true and without \"SYS_ADMIN\" capability is allowed",
         "template": "pod.yaml",
         "expected": "pass",
         "field_change_list": [
         ]
     },
     {
-        "name": "Deployment having container having securityContext.privileged set to true and with \"SYS_ADM\" capability is denied",
+        "name": "Deployment having container having securityContext.privileged set to true and with \"SYS_ADMIN\" capability is denied",
         "template": "deployment-for-list-items.yaml",
         "expected": "fail",
         "field_change_list": [
-            "spec.template.spec.containers.[0].securityContext.privileged=true"
+            "spec.template.spec.containers.[0].securityContext.privileged=true",
+            "spec.template.spec.containers.[0].securityContext.capabilities.add.[0]=SYS_ADMIN"
         ]
     },
     {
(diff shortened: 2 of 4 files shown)


controls/C-0057/policy.yaml
```diff
@@ -29,7 +29,7 @@ spec:
         (
           (!(has(container.securityContext.privileged)) || container.securityContext.privileged != true) &&
           (!(has(container.securityContext.capabilities)) || !(has(container.securityContext.capabilities.add)) ||
-          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADM')))
+          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADMIN')))
```
matthyx, line 32: Correct fix. Note this same expression is duplicated verbatim in `runtime-policies/privileged/policy.yaml` (lines 28, 37, 46), which still has `SYS_ADM` and is included in every release artifact via the `../../runtime-policies` base in all four `apis/*/kustomization.yaml`. Please fix it here too, otherwise the bypass still ships under `cluster-policy-deny-privileged-flag`.

controls/C-0057/tests.json
```diff
@@ -8,37 +8,40 @@
...
         "expected": "fail",
         "field_change_list": [
-            "spec.containers.[0].securityContext.privileged=true"
+            "spec.containers.[0].securityContext.privileged=true",
+            "spec.containers.[0].securityContext.capabilities.add.[0]=SYS_ADMIN"
         ]
     },
     {
-        "name": "Pod having container not having securityContext.privileged set to true and without \"SYS_ADM\" capability is allowed",
+        "name": "Pod having container not having securityContext.privileged set to true and without \"SYS_ADMIN\" capability is allowed",
```
matthyx, line 28: Non-blocking: this pass case uses `pod.yaml`, which has no `securityContext` at all, so it only exercises the `!has(container.securityContext)` short-circuit — nothing asserts that a container with a *benign* capability is still admitted. Consider adding one more case using the list-items template:

```json
{
    "name": "Pod having container with a non-SYS_ADMIN capability is allowed",
    "template": "pod-for-list-items.yaml",
    "expected": "pass",
    "field_change_list": [
        "spec.containers.[0].securityContext.capabilities.add.[0]=NET_ADMIN"
    ]
}
```

matthyx (changes requested):
Thanks for picking this up — the fix itself is correct and the test half is the valuable part. A couple of things need to land before this can merge, though, because the same bug is still live elsewhere in the repo.

**What I verified**

- `SYS_ADMIN` is the right spelling: Kubernetes `securityContext.capabilities.add` takes capability names without the `CAP_` prefix, and this repo already agrees with itself — `configuration/basic-control-configuration.yaml:19` lists `SYS_ADMIN` in `insecureCapabilities` (used by C-0046 / `runtime-policies/insecure-capabilities`). `SYS_ADM` matched nothing, so the capability arm of C-0057 was dead code.
- The `tests.json` change is what actually prevents this from regressing. The two "capability is denied" cases previously had an empty (or privileged-only) `field_change_list`, so the generated object never carried the capability and the fail result came entirely from `privileged: true` — nothing exercised the capability branch. They now set `...capabilities.add.[0]=SYS_ADMIN` against the `-for-list-items` templates, which already ship a populated `add:` list, so the index overwrite works and the assertions are real.
- CI is green (`test-all-policies`), DCO signed, no conflicts.

## Blockers

**1. The identical typo is still live in `runtime-policies/privileged/policy.yaml` (lines 28, 37, 46).**

That file is a near-verbatim copy of this policy (`cluster-policy-deny-privileged-flag`) and it ships: all four `apis/*/kustomization.yaml` list `../../runtime-policies` as a base, so every release artifact built by `.github/workflows/release.yaml` carries the un-fixed expression. Fixing only `controls/C-0057` leaves the bypass in shipped output under a different policy name. It's the same three-line change — please include it here so the two copies don't drift further.

**2. The user-facing doc still documents the wrong capability.**

`docs/policies-based-on-kubescape-controls/kubescape-c-0057-privileged-container-denied.md:24` reads:

> * If `securityContext.capabilities.add` does not contain `SYS_ADM` capability.

That page is hand-maintained and linked from the README control table, so after this merges the docs contradict the policy.

## Non-blocking (fine as follow-ups)

**3. No allow-side coverage for the capability branch.** Both `expected: "pass"` cases use `pod.yaml` / `deployment.yaml`, which have no `securityContext` at all — they only exercise the `!has(container.securityContext)` short-circuit. Nothing proves a container with a *benign* capability is still admitted, so an over-broad expression would slip through. One extra case would close it:

```json
{
    "name": "Pod having container with a non-SYS_ADMIN capability is allowed",
    "template": "pod-for-list-items.yaml",
    "expected": "pass",
    "field_change_list": [
        "spec.containers.[0].securityContext.capabilities.add.[0]=NET_ADMIN"
    ]
}
```

**4. The CronJob expression has no test at all.** All three expressions were touched, but `tests.json` only covers Pod and Deployment. Adding it needs a `cronjob-for-list-items.yaml` test resource (`scripts/change-yaml-field.py` can't create a new list element — `int('[0]')` raises, so it builds a dict and the apply fails), which is why I'd keep it out of this PR. Pre-existing gap, worth an issue.

**5. Exact string compare is bypassable — pre-existing, and it affects C-0046 too.** containerd's CRI normalizes with `"CAP_" + strings.ToUpper(c)`, so `add: ["sys_admin"]` is accepted by the API server, grants `CAP_SYS_ADMIN` at runtime, and sails past `cap != 'SYS_ADMIN'`. `add: ["ALL"]` likewise grants SYS_ADMIN and isn't matched. `cap.lowerAscii() != 'sys_admin'` plus an explicit `ALL` check would harden it, but since the same hole exists in `insecure-capabilities`/C-0046 it's a repo-wide follow-up, not something to bolt onto a typo fix.

**6. `initContainers` / `ephemeralContainers` are not covered** by C-0057, while C-0200 does cover them — a privileged init container is currently admitted. Pre-existing; also follow-up material.

**7. Nit:** `controls/C-0057/policy.yaml` still has no trailing newline (pre-existing).

Fix 1 and 2 and I'm happy to approve — ping me when they're pushed.

manmathbh:
@matthyx thanks for catching the runtime-policies clone and the docs! I've pushed an update addressing both of those blockers, and I also added the NET_ADMIN pass test you provided.
Let me know if we're good to go!

matthyx (approved):
Both blockers are addressed in `09a1e0b`, plus the suggested test. LGTM — approving.

- `runtime-policies/privileged/policy.yaml` — all three expressions (lines 28/37/46) now use `SYS_ADMIN`, so the release artifacts built from the `../../runtime-policies` kustomize base no longer ship the bypass.
- `docs/policies-based-on-kubescape-controls/kubescape-c-0057-privileged-container-denied.md:24` — now matches the policy.
- New pass case `Pod having container with a non-SYS_ADMIN capability is allowed` closes the allow-side gap: `pod-for-list-items.yaml` carries a populated `add:` list, so the `[0]=NET_ADMIN` overwrite produces `add: [NET_ADMIN]` and the case actually exercises the capability comparison rather than the `!has(securityContext)` short-circuit. `test-all-policies` is green on this commit, so it genuinely admitted.

The FOSSA contexts (Dependency Quality / License Compliance / Security Analysis) are still pending on the new commit — nothing to do with this change, they just need to re-run.

The earlier non-blocking items still stand as follow-ups, none of which should hold this up:

- No CronJob coverage for any of the three C-0057 expressions (needs a `cronjob-for-list-items.yaml` fixture).
- Exact-string compare is bypassable via `add: ["sys_admin"]` (containerd normalizes with `"CAP_" + ToUpper`) and `add: ["ALL"]` isn't matched — same gap in C-0046 / `insecure-capabilities`, so repo-wide.
- `initContainers` / `ephemeralContainers` aren't covered here, though C-0200 does cover them.
- `test-resources/*.yaml` fixtures still carry the invalid `SYS_ADM` string (also `pod-capabilities.yaml`, `deployment-with-common-label-{1,2}.yaml`). Harmless for the controls that use them, since those assert on "any capability added" rather than the name — but worth a cleanup pass so the typo doesn't get copied forward again.

Thanks for the quick turnaround.

PR #109, opened 2026-08-06
fix: prevent capability string casing and ALL wildcard bypass (Issue #106)

### LINKED ISSUE #106: Security Bypass: Exact-string compare for capabilities misses lowercase and ALL (C-0057, C-0046)
Currently, capability checks like `cap != 'SYS_ADMIN'` rely on exact string comparisons. This creates a bypass vulnerability because containerd normalizes capabilities at runtime by appending `CAP_` and converting to uppercase.

An attacker can bypass the policy by using:
* `add: ["sys_admin"]`
* `add: ["ALL"]` (which inherently grants `SYS_ADMIN`)

This gap affects both `C-0057` and `C-0046` (insecure-capabilities).

**Proposed Fix:**
Update the CEL expressions across these policies to convert the input to uppercase (or ignore case) and explicitly check for the `ALL` wildcard.

manmathbh:
Resolves #106.

Prevent two capability bypasses in C-0057, C-0046, and runtime-policies/privileged:

1. **Case bypass** - Convert both the incoming capability and the insecure capability list to uppercase using `upperAscii()` before comparison, so values like `sys_admin` or `net_raw` can no longer slip past.
2. **ALL wildcard bypass** - Explicitly deny the `ALL` capability string, which in Kubernetes grants every capability and was not previously caught.

### Files changed:
- `controls/C-0057/policy.yaml` - 3 expressions updated
- `controls/C-0046/policy.yaml` - 3 expressions updated
- `runtime-policies/privileged/policy.yaml` - 3 expressions updated
- `controls/C-0057/tests.json` - Added tests for lowercase `sys_admin` and `ALL`
- `controls/C-0046/tests.json` - Added tests for lowercase `net_raw` and `ALL`

## Summary by CodeRabbit

* **Bug Fixes**
  * Capability validation now treats capability names case-insensitively.
  * Pods, workloads, and CronJobs correctly reject the `ALL` capability.
  * Privileged-container checks consistently block both `SYS_ADMIN` and `ALL`, regardless of capitalization.

* **Tests**
  * Added coverage for lowercase capability names and blocked `ALL` capability scenarios.

### DIFF
--- a/controls/C-0046/policy.yaml
+++ b/controls/C-0046/policy.yaml
@@ -28,22 +28,28 @@ spec:
   validations:
     - expression: >
         object.kind != 'Pod' ||
-        object.spec.containers.all(container, params.settings.insecureCapabilities.all(insecureCapability,
+        object.spec.containers.all(container,
         !has(container.securityContext) || !has(container.securityContext.capabilities) || !has(container.securityContext.capabilities.add) ||
-        container.securityContext.capabilities.add.all(capability, capability != insecureCapability)
-        ))
+        (container.securityContext.capabilities.add.all(capability, capability.upperAscii() != 'ALL') &&
+        params.settings.insecureCapabilities.all(insecureCapability,
+        container.securityContext.capabilities.add.all(capability, capability.upperAscii() != insecureCapability.upperAscii())))
+        )
       message: "Pod has one or more containers with insecure capabilities! (see more at https://kubescape.io/docs/controls/c-0046/)"
     - expression: >
         ['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job'].all(kind, object.kind != kind) ||
-        object.spec.template.spec.containers.all(container, params.settings.insecureCapabilities.all(insecureCapability,
+        object.spec.template.spec.containers.all(container,
         !has(container.securityContext) || !has(container.securityContext.capabilities) || !has(container.securityContext.capabilities.add) ||
-        container.securityContext.capabilities.add.all(capability, capability != insecureCapability)
-        ))
+        (container.securityContext.capabilities.add.all(capability, capability.upperAscii() != 'ALL') &&
+        params.settings.insecureCapabilities.all(insecureCapability,
+        container.securityContext.capabilities.add.all(capability, capability.upperAscii() != insecureCapability.upperAscii())))
+        )
       message: "Workload has one or more containers with insecure capabilities! (see more at https://kubescape.io/docs/controls/c-0046/)"
     - expression: >
         object.kind != 'CronJob' ||
-        object.spec.jobTemplate.spec.template.spec.containers.all(container, params.settings.insecureCapabilities.all(insecureCapability,
+        object.spec.jobTemplate.spec.template.spec.containers.all(container,
         !has(container.securityContext) || !has(container.securityContext.capabilities) || !has(container.securityContext.capabilities.add) ||
-        container.securityContext.capabilities.add.all(capability, capability != insecureCapability)
-        ))
+        (container.securityContext.capabilities.add.all(capability, capability.upperAscii() != 'ALL') &&
+        params.settings.insecureCapabilities.all(insecureCapability,
+        container.securityContext.capabilities.add.all(capability, capability.upperAscii() != insecureCapability.upperAscii())))
+        )
       message: "CronJob has one or more containers with insecure capabilities! (see more at https://kubescape.io/docs/controls/c-0046/)"
--- a/runtime-policies/privileged/policy.yaml
+++ b/runtime-policies/privileged/policy.yaml
@@ -25,7 +25,7 @@ spec:
         (
           (!(has(container.securityContext.privileged)) || container.securityContext.privileged != true) &&
           (!(has(container.securityContext.capabilities)) || !(has(container.securityContext.capabilities.add)) ||
-          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADMIN')))
+          container.securityContext.capabilities.add.all(cap, cap.upperAscii() != 'SYS_ADMIN' && cap.upperAscii() != 'ALL')))
         )
       message: "Pod has one or more privileged container.(see more at https://kubescape.io/docs/controls/c-0057/)"
     - expression: >
@@ -34,7 +34,7 @@ spec:
         (
           (!(has(container.securityContext.privileged)) || container.securityContext.privileged != true) &&
           (!(has(container.securityContext.capabilities)) || !(has(container.securityContext.capabilities.add)) ||
-          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADMIN')))
+          container.securityContext.capabilities.add.all(cap, cap.upperAscii() != 'SYS_ADMIN' && cap.upperAscii() != 'ALL')))
         )
       message: "Workloads has one or more privileged container.(see more at https://kubescape.io/docs/controls/c-0057/)"
     - expression: >
@@ -43,6 +43,6 @@ spec:
         (
           (!(has(container.securityContext.privileged)) || container.securityContext.privileged != true) &&
           (!(has(container.securityContext.capabilities)) || !(has(container.securityContext.capabilities.add)) ||
-          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADMIN')))
+          container.securityContext.capabilities.add.all(cap, cap.upperAscii() != 'SYS_ADMIN' && cap.upperAscii() != 'ALL')))
         )
       message: "CronJob has one or more privileged container.(see more at https://kubescape.io/docs/controls/c-0057/)"
--- a/controls/C-0046/tests.json
+++ b/controls/C-0046/tests.json
@@ -33,5 +33,21 @@
         "expected": "fail",
         "field_change_list": [
         ]
+    },
+    {
+        "name": "Pod having container with lowercase \"net_raw\" capability is blocked",
+        "template": "pod-for-list-items.yaml",
+        "expected": "fail",
+        "field_change_list": [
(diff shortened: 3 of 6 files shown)


controls/C-0046/policy.yaml
```diff
@@ -30,20 +30,20 @@ spec:
         object.kind != 'Pod' ||
         object.spec.containers.all(container, params.settings.insecureCapabilities.all(insecureCapability,
         !has(container.securityContext) || !has(container.securityContext.capabilities) || !has(container.securityContext.capabilities.add) ||
-        container.securityContext.capabilities.add.all(capability, capability != insecureCapability)
+        container.securityContext.capabilities.add.all(capability, capability.upperAscii() != insecureCapability.upperAscii() && capability.upperAscii() != 'ALL')
```
matthyx: **Blocker** — same on lines 40 and 47, and in `runtime-policies/insecure-capabilities/policy.yaml` once that file is updated.

`capability.upperAscii() != 'ALL'` is evaluated *inside* `params.settings.insecureCapabilities.all(...)`, so the `ALL` denial only exists while that list is non-empty. `insecureCapabilities` is operator-configurable (`configuration/basic-control-configuration.yaml`) and in CEL `[].all(...)` is vacuously `true` — so a ControlConfiguration with `insecureCapabilities: []` (or omitting the key) makes the whole expression pass and re-opens the exact `ALL` bypass this PR closes.

Hoist the `ALL` check out of the params loop so it stands on its own, e.g.:

```yaml
        object.kind != 'Pod' ||
        object.spec.containers.all(container,
        !has(container.securityContext) || !has(container.securityContext.capabilities) || !has(container.securityContext.capabilities.add) ||
        (container.securityContext.capabilities.add.all(capability, capability.upperAscii() != 'ALL') &&
        params.settings.insecureCapabilities.all(insecureCapability,
        container.securityContext.capabilities.add.all(capability, capability.upperAscii() != insecureCapability.upperAscii())))
        )
```

Side benefit: this stops re-evaluating the `ALL` comparison once per configured capability, which keeps the VAP cost estimate down.

C-0057 and `runtime-policies/privileged` don't have this problem — the comparison there is against a literal, not a params list.

runtime-policies/privileged/policy.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         (
           (!(has(container.securityContext.privileged)) || container.securityContext.privileged != true) &&
           (!(has(container.securityContext.capabilities)) || !(has(container.securityContext.capabilities.add)) ||
-          container.securityContext.capabilities.add.all(cap, cap != 'SYS_ADMIN')))
+          container.securityContext.capabilities.add.all(cap, cap.upperAscii() != 'SYS_ADMIN' && cap.upperAscii() != 'ALL')))
         )
```
matthyx, line 29: This change is correct, but note that nothing in `runtime-policies/` is covered by CI — `scripts/run-all-control-tests.sh` iterates `controls/` only, so `test-all-policies` being green says nothing about this file. The expression is identical to the C-0057 one that *is* tested, so I'm satisfied it works; flagging the coverage gap rather than asking for a change here.

The actionable part is the sibling file: `runtime-policies/insecure-capabilities/policy.yaml` is the C-0046 mirror and was not updated — see the review summary.

matthyx (changes requested):
Thanks for taking #106 on — `upperAscii()` normalization plus an explicit `ALL` denial is the right fix, the CEL string extension is available in K8s VAP, and `test-all-policies` is green. Two things to sort out before this can go in.

## 🔴 Blocker — the C-0046 runtime twin was missed

`runtime-policies/insecure-capabilities/policy.yaml` (`cluster-policy-deny-insecure-capabilities`) is the runtime mirror of C-0046 and still carries the unpatched comparison on all three branches:

```yaml
container.securityContext.capabilities.add.all(capability, capability != insecureCapability)
```

You correctly updated `runtime-policies/privileged` (the C-0057 mirror), so this looks like a plain oversight — but as it stands, anyone deploying `runtime-policies/kustomization.yaml` still has both bypasses from #106 (`net_raw` and `ALL`) wide open. Please apply the same change there.

On CodeRabbit's "Out of Scope Changes" warning about `runtime-policies/privileged`: I disagree with it. Keeping the runtime mirrors in sync with the controls is the right call — the fix is to update the *other* mirror too, not to drop this one.

## 🟠 Should fix — the `ALL` denial is conditional on `insecureCapabilities` being non-empty

In C-0046 (and its runtime twin) the new `capability.upperAscii() != 'ALL'` clause sits inside `params.settings.insecureCapabilities.all(...)`. That list is operator-configurable, and `[].all(...)` is vacuously `true` in CEL — so `insecureCapabilities: []` disables the `ALL` check along with the rest of the control. The `ALL` denial shouldn't depend on the params list at all. Suggested rewrite inline.

## Notes — not blocking

- **New tests are Pod-only.** All four added cases use `pod-for-list-items.yaml`, so the workload-template and CronJob branches you edited in both controls are unexercised for the new behavior. One `deployment-for-list-items.yaml` case per control would be cheap insurance (same point CodeRabbit raised).
- **`runtime-policies/` has no CI.** `scripts/run-all-control-tests.sh` only walks `controls/`, so the `privileged` edit is unverified by the pipeline. Low risk since it mirrors a tested expression — noting it as a gap, not asking for it here.
- **Follow-ups, out of scope for #106** (worth separate issues):
  - `controls/C-0199` has the identical class of bug — `cap != 'NET_RAW' && cap != 'ALL'` and `drop.exists(c, c == 'NET_RAW' || c == 'ALL')` are both case-sensitive.
  - C-0046 and C-0057 evaluate only `spec.containers`; `initContainers` and `ephemeralContainers` are never checked, so `ALL` in an initContainer still gets through. C-0199/C-0200/C-0201 already cover all three container lists.

Happy to approve once `runtime-policies/insecure-capabilities/policy.yaml` is updated and the `ALL` check is hoisted out of the params loop.

manmathbh:
@matthyx , I've hoisted the ALL check completely out of the params loop in both C-0046 and its runtime mirror so it evaluates independently. Both blockers should be resolved now!

I also noted your points on C-0199 and the container scope gaps.
PTAL.

matthyx (approved):
Both blockers addressed — approving. 🎉

- ✅ `runtime-policies/insecure-capabilities/policy.yaml` is now patched, so the C-0046 runtime mirror no longer carries the bypass.
- ✅ The `ALL` denial is hoisted out of the `params.settings.insecureCapabilities.all(...)` loop in all three branches of both C-0046 and its runtime twin, so it no longer depends on the operator-supplied list being non-empty.

## Verification

CI on `f52045e` was **cancelled** mid-run (`test-all-policies`, run 31118126545 — the job was killed at ~22 min, no logs archived), so nothing green covers the restructured C-0046 expression yet. I've re-triggered that run (attempt 2), but since the restructure is non-trivial I also evaluated the expressions directly with `cel-go` v0.30.0 (`ext.Strings()` enabled, `object`/`params` as dyn) — all four expressions compile, and 26/26 behavioral cases match expectations:

| case | expected | result |
|---|---|---|
| Pod `SYS_ADM`, default params | admit | ✅ |
| Pod no `securityContext` | admit | ✅ |
| Pod `NET_RAW` / `net_raw` / `NeT_rAw` | deny | ✅ |
| Pod `ALL` / `all` | deny | ✅ |
| Pod `["SYS_ADM","ALL"]` (ALL not first) | deny | ✅ |
| **Pod `ALL` with `insecureCapabilities: []`** | **deny** | ✅ ← the hoist fix |
| Pod `SYS_ADM` with `insecureCapabilities: []` | admit | ✅ |
| Deployment / CronJob: `net_raw`, `ALL`, `ALL` + empty params | deny | ✅ |
| Deployment / CronJob: benign cap | admit | ✅ |
| Kind guards (Pod expr vs Deployment object, and vice versa) | admit | ✅ |
| C-0057/privileged: `SYS_ADMIN`, `sys_admin`, `ALL`, `all` | deny | ✅ |
| C-0057/privileged: `NET_ADMIN`, no `securityContext` | admit | ✅ |

Parenthesization on the restructured expression is correct — `all(container, A \|\| B \|\| C \|\| (D && E))` — and no benign-capability case regressed.

**Please confirm `test-all-policies` goes green on the re-run before merging** — my approval covers the code, not the pipeline.

## Follow-ups (not blocking this PR)

- New tests still only exercise the Pod branch; the workload-template and CronJob branches you edited are unexercised for the new behavior. Worth a `deployment-for-list-items.yaml` case per control at some point.
- Nothing under `runtime-policies/` is covered by CI — `scripts/run-all-control-tests.sh` walks `controls/` only. Both runtime policies here are verified only by the cel-go run above.
- Separate issues, outside #106's scope: `controls/C-0199` has the identical case-sensitivity bug (`cap != 'NET_RAW' && cap != 'ALL'`, `drop.exists(c, c == 'NET_RAW' || c == 'ALL')`), and C-0046/C-0057 evaluate only `spec.containers`, so `ALL` in an `initContainer` or `ephemeralContainer` still gets through.

Nice work on the turnaround.

matthyx:
Heads up on the red `test-all-policies` — it's **not** your change. GitHub Actions is in an active major outage ([incident opened 15:22 UTC](https://www.githubstatus.com/), still investigating; Actions currently `major_outage`).

Both failures on `f52045e` are outage symptoms, and neither ever reached the test step:

- **Attempt 1** — job sat for 22 min and was cancelled with no steps recorded and no logs archived.
- **Attempt 2** (my re-run) — died in `Set up job`, before checkout: `Failed to resolve action download info. Error: Service Unavailable`, after two retries.

The last green run (`0a24aa8`, 14:29 UTC) predates the incident.

Nothing to do here — no need to push an empty commit or chase a fix. Once GitHub recovers, re-run the workflow and it should go green. My approval stands; the policy logic itself I verified independently with cel-go (26/26 cases) in the review above.

PR #112, opened 2026-08-06
fix: add missing controlId labels to C-0012 and C-0013

Sanchit2662:
C-0012 and C-0013 were the only two policies in the repo without a `controlId` label. They were also the only two without the `controlUrl` annotation, so I added both to keep them in line with the other 38.

Why it matters: kubescape indexes the synced policy bundle by the `controlId` label. A policy with no label still gets indexed by name, but it never gets attached to a control, so the scan path just skips it. Same on the enforcement status side, those two controls would keep showing as not enforced even with the policy installed and bound in the cluster.

Nothing under `spec` changed, this is metadata only. `kubectl kustomize controls/` still builds and I ran the C-0012 and C-0013 test suites locally against a 1.31 cluster, both green.

Two unrelated things I noticed while doing this, I did not touch them here:

* the README row and the doc filename for C-0013 both say `kubescape-c-0013-deny-if-container-runs-as-root`, but the policy's real `metadata.name` is `kubescape-c-0013-deny-resources-with-capability-to-run-as-root`
* `docs/policies-based-on-kubescape-controls/kubescape-c-0013-deny-if-container-runs-as-root.md` is a byte for byte copy of the C-0017 doc, so C-0013 has no actual doc page right now

Can send those as a follow up if you want them fixed.

## Summary by CodeRabbit

* **Documentation**
  * Added control identifiers and reference links to two admission policies.
  * Improved policy metadata for easier traceability and navigation.
  * Policy validation behavior remains unchanged.

### DIFF
--- a/controls/C-0012/policy.yaml
+++ b/controls/C-0012/policy.yaml
@@ -2,6 +2,10 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingAdmissionPolicy
 metadata:
   name: "kubescape-c-0012-deny-resources-with-sensitive-information-in-environment-variables"
+  labels:
+    controlId: "C-0012"
+  annotations:
+    controlUrl: "https://kubescape.io/docs/controls/c-0012/"
 spec:
   failurePolicy: Fail
   paramKind:
--- a/controls/C-0013/policy.yaml
+++ b/controls/C-0013/policy.yaml
@@ -2,6 +2,10 @@ apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingAdmissionPolicy
 metadata:
   name: "kubescape-c-0013-deny-resources-with-capability-to-run-as-root"
+  labels:
+    controlId: "C-0013"
+  annotations:
+    controlUrl: "https://kubescape.io/docs/controls/c-0013/"
 spec:
   failurePolicy: Fail
   matchConstraints:


matthyx (approved):
LGTM — no blockers. Metadata-only, and it closes a real gap.

**What I checked**

- C-0012 and C-0013 were indeed the only two of the 40 policies missing `controlId` / `controlUrl` — the other 38 have both, and now all 40 do.
- Format matches the existing convention exactly: same key names, same `labels` → `annotations` ordering under `metadata`, quoted values, lowercase slug and trailing slash in the URL. Nothing to nit.
- `kubectl kustomize controls/` builds clean on the PR head — 40 `ValidatingAdmissionPolicy` docs, with `controlId` present on C-0012 and C-0013.
- `spec` is untouched on both files (diff is +8 metadata lines, 0 deletions), so admission behaviour is unchanged.
- Both `controlUrl` targets resolve (`c-0012/` and `c-0013/` → 200), and `C-0012` / `C-0013` are valid Kubernetes label values.

**Non-blocking notes**

1. CodeRabbit's inline comment on `controls/C-0013/policy.yaml` is about the C-0013 doc page, not about this diff. It's a pre-existing docs bug you already called out in the description — leaving it out of a metadata-only PR is the right call.

2. On those two follow-ups, both confirmed:
   - `docs/policies-based-on-kubescape-controls/kubescape-c-0013-deny-if-container-runs-as-root.md` is byte-for-byte identical to the C-0017 doc (`diff` returns nothing), so C-0013 genuinely has no doc page.
   - The README row is generated from `metadata.name` by `scripts/get-control-table.py`, so re-running that script fixes the stale filename/link half automatically. The doc page itself still has to be written by hand.

   A follow-up PR doing exactly that would be welcome — please keep it separate from this one.

3. FYI, `main-policy-automation` has no run against head `b38dd7d` — it hasn't executed on this PR. Zero risk here given nothing under `spec` changed, but worth knowing the policy test suite isn't part of the green checks you're seeing.

Approving.

Issue #99, opened 2026-07-28
Controls picked for the Rego to CEL migration [LFX'26]


Sanchit2662:
## Overview
kubescape/kubescape#2002 asks for existing regolibrary controls to be migrated but does not say which ones, so this is my proposal. 20 controls total. All of them are ready to start.

## How I picked them

Three things, in this order:

1. The control has to be convertible at all. No host scanner, no reading a second resource, no OPA builtin that CEL does not have.
2. Between them they should exercise the parts of the engine we built, so the migration actually tests it. String matching, numeric parsing, list iteration, map iteration, multi kind matching, and one that is expensive enough to stress the cost budget.
3. Reach. I checked baseScore and how many frameworks each control appears in, so we are not converting things nobody scans.

I also ran every expression below through `EvaluateOnObject` against real objects before listing it, so these are checked and not just guesses. That is also how I found the problem in the next section.

## One thing worth knowing first

I originally had C-0190 and C-0259 in this list. Both turned out to be already done.

In regolibrary one Rego rule can belong to several controls, so checking "is this control ID in cel-admission-library" gives the wrong answer. Checking by rule name instead: C-0190 shares `automount-service-account` with C-0034, and C-0034's policy is already exactly the expression C-0190 needs. C-0259 shares `rule-credentials-in-env-var` with C-0012, and C-0012 already covers it with the same params plus ConfigMaps on top.

So I swapped them for C-0225 and C-0296.

Small related thing: C-0012 and C-0013 are in the bundle but have no `controlId` label, so the kubescape loader cannot find them. Probably why the C-0259 overlap was not obvious. Happy to send a separate PR for that.

## The list (13)

- [x] **C-0275** Pod with `hostPID: true`. One bool, three kind shapes. Simplest thing in the list, good first PR to shake out the five file process and the release plus pin bump loop.
- [x] **C-0276** Same as above for `hostIPC`. Same PR as C-0275, it is a copy paste.
- [x] **C-0263** Ingress with no `spec.tls`. First non workload kind, and a one liner. Needs a new `test-resources/ingress.yaml`.
- [x] **C-0292** Workload running the community ingress-nginx controller, which hits end of life in March 2026. String matching on the image. Good reach, 4 frameworks.
[...]

PR #113, opened 2026-08-07
feat: add C-0275 and C-0276 for host PID and IPC namespace sharing

Sanchit2662:

These are the first two controls off the migration list. C-0275 is `host-pid-privileges` in regolibrary and C-0276 is `host-ipc-privileges`.

At first glance this looks like a duplicate of C-0038, which already checks hostPID and hostIPC together. It is not. These are two separate CIS items with their own control IDs and their own frameworks (cis-v1.10.0 and cis-v1.12.0), and kubescape ties a policy back to a control through the `controlId` label, which only holds one value. So each control needs its own file. C-0276 is basically C-0275 with the field name swapped.

I went with the newer `variables:` style from the conversion guide. There is no container walk needed here, so the variable just grabs the pod spec for whatever kind is coming in:

```
object.kind == 'Pod' ? object.spec
  : object.kind in ['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job'] ? object.spec.template.spec
  : object.kind == 'CronJob' ? object.spec.jobTemplate.spec.template.spec
  : {}
```

and then the actual check is one line, `!has(variables.podSpec.hostPID) || variables.podSpec.hostPID == false`.

The `!has(...)` bit matters more than it looks. Both fields are omitempty, so a normal workload does not carry them at all, and without that guard every ordinary pod would fail to evaluate instead of just passing.

One difference from the Rego that I want to say out loud. The Rego rules emit a `deletePath` for `spec.hostPID` and `spec.hostIPC`, and the CEL side is going to end up producing a fix that writes `false` instead. Both fields default to false so you land in the same place either way, and C-0038 already behaves like this.

For tests I wrote 14 cases each. Absent, `false` and `true` on Pod, Deployment and CronJob, plus ReplicaSet, DaemonSet, StatefulSet and Job on the `true` case. Each suite also has a case where the other field is set to true and is expected to pass, so C-0275 cannot quietly start catching hostIPC and the other way around. All 28 pass for me on a local 1.31 cluster.


matthyx (approved):
LGTM! Added C-0275 (deny host PID namespace) and C-0276 (deny host IPC namespace) with complete policy definitions, comprehensive tests for all supported workload kinds, documentation, and kustomization updates. All CI checks pass.

PR #114, opened 2026-08-08
feat: add C-0263, C-0292 and C-0081

Sanchit2662:

Three more off the list. C-0263 is `ingress-no-tls` in regolibrary, C-0292 is `detect-nginx-ingress-controller-eol` and C-0081 is `CVE-2022-24348`.

## C-0263, Ingress with no TLS

First control here that matches something other than a workload, so matchConstraints is `networking.k8s.io/v1` ingresses. The check itself is one line, `has(object.spec.tls) && object.spec.tls.size() > 0`. I also added `test-resources/ingress.yaml`, there was no Ingress fixture in the repo yet.

One small difference from the Rego that I want to be upfront about. The Rego does `not ingress.spec.tls`, so an Ingress with `tls: []` passes there. Mine fails it, because an empty list configures no TLS at all. I think failing is the right call but it is a real behaviour change, so say if you disagree and I will drop the size check.

## C-0292, community ingress-nginx EOL

Deployment, DaemonSet and StatefulSet only. The Rego does not cover Pod, Job or CronJob so I did not widen it either, even though most policies in here bind all seven kinds. Widening would change which resources fail.

I kept the three exact substrings from the Rego rather than simplifying to something like `contains('nginx')`. That is on purpose. The comment in the Rego says it only wants the community project, not the NGINX Inc commercial images, which have their own support lifecycle. There is a test for that.

Main containers only, not init or ephemeral, same as the Rego. An ingress controller sitting in an init container is not actually running.

## C-0081, argocd CVE-2022-24348

Deployment only, again like the Rego. This is the fiddly one. It pulls the tag out of the image, splits it on dots, and needs three numeric parts before it can compare anything.

Two guards that are not optional:

* `size() != 3` has to come first, or a tag like `v2.1` makes `parts[2]` raise
* `matches('^[0-9]+$')` has to come before `int()`, or a tag like `v2.1.0-rc1` makes `int()` raise

A raise is not a pass. It is an eval error, and the scanner reports that as skipped, so getting the order wrong would quietly lose the control instead of failing loudly. Both cases have a test.

The `[ ... ].all(parts, ...)` wrapper looks odd, I know. CEL has no let binding, and without it I would be repeating that same long split eleven times in one expression. Wrapping the value in a one element list is the usual way to give it a name. Happy to write it out the long way instead if you would rather not have that pattern in the repo.

## Tests

4 for C-0263, 9 for C-0292, 11 for C-0081. The C-0081 ones sit on both boundaries, 2.1.9 and 2.2.4 pass while 2.1.8 and 2.2.3 fail, plus the two guard tags and a plain nginx image. All 24 pass for me on a local 1.31 cluster.

## Summary by CodeRabbit

* **New Features**
  * Added policies blocking vulnerable Argo CD versions affected by CVE-2022-24348.
  * Added enforcement against Ingress resources without TLS configuration.
  * Added detection of end-of-life community ingress-nginx controllers across supported workload types, while allowing commercial NGINX images.
* **Documentation**
  * Added policy documentation covering scope, severity, configuration, and deployment guidance.
* **Tests**
  * Added comprehensive validation scenarios for vulnerable images, TLS settings, containers, and workload types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,9 @@ kubectl -n policy-example run nginx --image=nginx --restart=Never
 | [C-0271](https://kubescape.io/docs/controls/c-0271/) | Ensure memory limits are set | [kubescape-c-0271-deny-resources-with-memory-limit-not-set](/docs/policies-based-on-kubescape-controls/kubescape-c-0271-deny-resources-with-memory-limit-not-set.md) | [memoryLimitMin](https://kubescape.io/docs/frameworks-and-controls/configuring-controls/#memory_limit_min) |
 | [C-0275](https://kubescape.io/docs/controls/c-0275/) | Minimize the admission of containers wishing to share the host process ID namespace | [kubescape-c-0275-deny-resources-sharing-host-pid-namespace](/docs/policies-based-on-kubescape-controls/kubescape-c-0275-deny-resources-sharing-host-pid-namespace.md) | not configurable |
 | [C-0276](https://kubescape.io/docs/controls/c-0276/) | Minimize the admission of containers wishing to share the host IPC namespace | [kubescape-c-0276-deny-resources-sharing-host-ipc-namespace](/docs/policies-based-on-kubescape-controls/kubescape-c-0276-deny-resources-sharing-host-ipc-namespace.md) | not configurable |
+| [C-0081](https://kubescape.io/docs/controls/c-0081/) | CVE-2022-24348 Argo CD directory traversal | [kubescape-c-0081-deny-vulnerable-argocd-versions](/docs/policies-based-on-kubescape-controls/kubescape-c-0081-deny-vulnerable-argocd-versions.md) | not configurable |
+| [C-0263](https://kubescape.io/docs/controls/c-0263/) | Ingress uses TLS | [kubescape-c-0263-deny-ingress-without-tls](/docs/policies-based-on-kubescape-controls/kubescape-c-0263-deny-ingress-without-tls.md) | not configurable |
+| [C-0292](https://kubescape.io/docs/controls/c-0292/) | Nginx Ingress Controller End of Life | [kubescape-c-0292-deny-nginx-ingress-controller-eol](/docs/policies-based-on-kubescape-controls/kubescape-c-0292-deny-nginx-ingress-controller-eol.md) | not configurable |
 
 ## Testing Policies
 
--- a/controls/C-0081/policy.yaml
+++ b/controls/C-0081/policy.yaml
@@ -1,0 +1,47 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "kubescape-c-0081-deny-vulnerable-argocd-versions"
+  labels:
+    controlId: "C-0081"
+  annotations:
+    controlUrl: "https://kubescape.io/docs/controls/c-0081/"
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments"]
+  variables:
+    - expression: |
+        object.kind == 'Deployment'
+          ? object.spec.template.spec.containers + (has(object.spec.template.spec.initContainers) ? object.spec.template.spec.initContainers : [])
+          : []
+      name: containers
+  validations:
+    # Argo CD fixed CVE-2022-24348 in 2.1.9 and 2.2.4. Vulnerable: any 1.x, any 2.0.x,
+    # 2.1.x below 9 and 2.2.x below 4.
+    # argocd has to be a whole repository segment, otherwise an unrelated image such as
+    # myargocd:v1.8.7 would be denied.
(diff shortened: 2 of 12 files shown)


controls/C-0081/policy.yaml
```diff
@@ -0,0 +1,45 @@
...
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments"]
+  variables:
+    - expression: |
+        object.kind == 'Deployment'
+          ? object.spec.template.spec.containers
```
matthyx: Missing `initContainers` in `variables.containers`. Please update to: `object.spec.template.spec.containers + (has(object.spec.template.spec.initContainers) ? object.spec.template.spec.initContainers : [])` to ensure init containers are also checked.

controls/C-0081/policy.yaml
```diff
@@ -0,0 +1,45 @@
...
+  validations:
+    # Argo CD fixed CVE-2022-24348 in 2.1.9 and 2.2.4. Vulnerable: any 1.x, any 2.0.x,
+    # 2.1.x below 9 and 2.2.x below 4.
+    # The single-element list is only there to bind the split result to a name, CEL has
+    # no let binding and repeating the split eleven times is unreadable.
+    # size() and matches() must both be checked before int(), because int() raises on a
+    # tag like v2.1.0-rc1 and a raise is an eval error, not a pass.
+    - expression: >
+        variables.containers.all(container,
+          !container.image.contains('argocd:v') ||
```
matthyx: Substrings like `argocd:v` cause false positives for images such as `myargocd:v1.8.7`. Please refine image matching to check `container.image.contains('/argocd:v') || container.image.startsWith('argocd:v')`.

controls/C-0292/policy.yaml
```diff
@@ -0,0 +1,33 @@
...
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments","daemonsets","statefulsets"]
+  variables:
+    - expression: |
+        object.kind in ['Deployment','DaemonSet','StatefulSet']
+          ? object.spec.template.spec.containers
```
matthyx: Missing `initContainers` in `variables.containers`. Please update to: `object.spec.template.spec.containers + (has(object.spec.template.spec.initContainers) ? object.spec.template.spec.initContainers : [])` so init containers in workloads are validated.

README.md
```diff
@@ -88,6 +88,9 @@ kubectl -n policy-example run nginx --image=nginx --restart=Never
...
 | [C-0270](https://kubescape.io/docs/controls/c-0270/) | Ensure CPU limits are set | [kubescape-c-0270-deny-resources-with-cpu-limit-not-set](/docs/policies-based-on-kubescape-controls/kubescape-c-0270-deny-resources-with-cpu-limit-not-set.md) | [cpuLimitMin](https://kubescape.io/docs/frameworks-and-controls/configuring-controls/#cpu_limit_min) |
 | [C-0271](https://kubescape.io/docs/controls/c-0271/) | Ensure memory limits are set | [kubescape-c-0271-deny-resources-with-memory-limit-not-set](/docs/policies-based-on-kubescape-controls/kubescape-c-0271-deny-resources-with-memory-limit-not-set.md) | [memoryLimitMin](https://kubescape.io/docs/frameworks-and-controls/configuring-controls/#memory_limit_min) |
+| [C-0081](https://kubescape.io/docs/controls/c-0081/) | CVE-2022-24348-argocddirtraversal | [kubescape-c-0081-deny-vulnerable-argocd-versions](/docs/policies-based-on-kubescape-controls/kubescape-c-0081-deny-vulnerable-argocd-versions.md) | not configurable |
```
matthyx: Please update table name to `CVE-2022-24348 Argo CD directory traversal` to match the control title format used across the README table and doc.

matthyx (changes requested):
Thank you for contributing C-0263, C-0292, and C-0081! C-0263 looks great. Please address the following blocking items in C-0081, C-0292, and README.md before merge:

1. **Include `initContainers` in container collections**: In `C-0081` and `C-0292`, `variables.containers` currently only extracts `object.spec.template.spec.containers`. Workloads running vulnerable images inside `initContainers` would bypass validation. Please update `variables.containers` to include `initContainers` (e.g. `object.spec.template.spec.containers + (has(object.spec.template.spec.initContainers) ? object.spec.template.spec.initContainers : [])`).
2. **Prevent false positives in Argo CD image matching**: In `C-0081/policy.yaml`, `container.image.contains('argocd:v')` matches any image containing `argocd:v` as a substring (such as `myargocd:v1.8.7`). Please refine the match to check for `/argocd:v` or starting with `argocd:v` (e.g. `container.image.contains('/argocd:v') || container.image.startsWith('argocd:v')`).
3. **Fix C-0081 title in README.md**: Update the control title in `README.md` to `CVE-2022-24348 Argo CD directory traversal` for consistency with the doc title and other entries.

matthyx (approved):
LGTM! All requested changes (initContainers support, image matching precision, and README formatting) have been addressed and all tests are passing. Good to merge!

PR #115, opened 2026-08-10
feat: add C-0207, C-0262 and C-0231

Sanchit2662:

Three more. C-0207 is `rule-secrets-in-env-var` in regolibrary, C-0262 is `anonymous-access-enabled` and C-0231 is `ensure-https-loadbalancers-encrypted-with-tls-aws`.

Also adds two fixtures, `test-resources/role-binding.yaml` and `test-resources/cluster-role-binding.yaml`, since the repo had Role and ClusterRole but no bindings.

## C-0207, Secrets injected as env vars

First one here with a nested list walk, containers and then env inside each container. Both `has()` guards are load bearing: `env` is optional on a container and `valueFrom` is optional on an env entry, and `has(env.valueFrom.secretKeyRef)` on its own does not cover a missing `valueFrom`.

I walk init containers, and ephemeral containers for Pods, which the Rego does not. That makes the CEL stricter than the Rego. I think it is right, an init container leaking a secret into its environment is the same problem, and it matches what the rest of the library does now. Flagging it because it is a real difference.

Tests cover secretKeyRef, and also configMapKeyRef, fieldRef and a plain value, so it is clear the policy only reacts to Secrets.

## C-0262, anonymous RoleBindings

RoleBinding and ClusterRoleBinding, checking `subjects[*].name` against `system:anonymous` and `system:unauthenticated`. `subjects` is optional so it needs a guard, and a binding with no subjects passes since it grants nothing.

The Rego does `endswith(rolebinding.kind, "Binding")` over whatever is in the input. I did not try to reproduce that on `object.kind`. matchConstraints is the right equivalent here and it is tighter.

There is a test for `system:authenticated` passing, so the two anonymous names do not accidentally grow into a prefix match.

## C-0231, LoadBalancer on 443 with no TLS cert

Only control in the set that iterates map keys instead of list elements, and it is a key prefix match rather than an exact lookup, because the AWS annotation family has suffixed variants. There is a test for a suffixed key.

I put the annotation check first in the disjunction on purpose. It reads slightly odd, but the remediation path we derive comes from the first object field walked, and the finding is really about the missing annotation, not about `spec.type`.

`spec.type` and `spec.ports` are both guarded even though the apiserver defaults `spec.type` before admission. A Service read straight out of a YAML file during a scan does not go through defaulting, and an ExternalName or headless Service has no ports.

One thing worth knowing: this control ships a `filter.rego` narrowing to `spec.type == "LoadBalancer"`. matchConstraints can only filter on kind, so the CEL version evaluates every Service. The verdicts are the same, but the risk score denominator will not be. There is a deliberate ClusterIP test case and a NodePort one so that shows up here rather than surprising someone in a parity run later.

## Summary by CodeRabbit

- **New Features**
  - Added Kubernetes admission controls to prevent Secrets from being exposed through container environment variables.
  - Added protection requiring AWS TLS certificates for `LoadBalancer` Services serving port 443.
  - Added protection against anonymous or unauthenticated subjects in RoleBindings and ClusterRoleBindings.
  - Registered all three controls for deployment and included coverage for workloads, Services, and RBAC bindings.

- **Documentation**
  - Added usage guidance, policy details, severity information, and remediation recommendations for each control.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,9 @@ kubectl -n policy-example run nginx --image=nginx --restart=Never
 | [C-0081](https://kubescape.io/docs/controls/c-0081/) | CVE-2022-24348 Argo CD directory traversal | [kubescape-c-0081-deny-vulnerable-argocd-versions](/docs/policies-based-on-kubescape-controls/kubescape-c-0081-deny-vulnerable-argocd-versions.md) | not configurable |
 | [C-0263](https://kubescape.io/docs/controls/c-0263/) | Ingress uses TLS | [kubescape-c-0263-deny-ingress-without-tls](/docs/policies-based-on-kubescape-controls/kubescape-c-0263-deny-ingress-without-tls.md) | not configurable |
 | [C-0292](https://kubescape.io/docs/controls/c-0292/) | Nginx Ingress Controller End of Life | [kubescape-c-0292-deny-nginx-ingress-controller-eol](/docs/policies-based-on-kubescape-controls/kubescape-c-0292-deny-nginx-ingress-controller-eol.md) | not configurable |
+| [C-0207](https://kubescape.io/docs/controls/c-0207/) | Prefer using secrets as files over secrets as environment variables | [kubescape-c-0207-deny-secrets-as-environment-variables](/docs/policies-based-on-kubescape-controls/kubescape-c-0207-deny-secrets-as-environment-variables.md) | not configurable |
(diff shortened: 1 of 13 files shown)


matthyx (approved):
LGTM! Addressed all review findings for C-0207, C-0231, and C-0262:
- C-0207: Added validation and tests for `container.envFrom` with `secretRef`.
- C-0231: Required exact annotation key `service.beta.kubernetes.io/aws-load-balancer-ssl-cert` with a non-empty string value, updated test cases accordingly, and cleaned up policy description.
- C-0262: Restricted `system:anonymous` to `kind: User` and `system:unauthenticated` to `kind: Group`.

All 38 policy tests pass.

PR #117, opened 2026-08-12
add C-0234, C-0295 and C-0212

Sanchit2662:

Three more. C-0234 is `ensure-external-secrets-storage-is-in-use` in regolibrary, C-0295 is `duplicate-env-var`, and C-0212 is the 17 `*-in-default-namespace` rules rolled into one policy.

Also adds ten fixtures under `test-resources/`. C-0212 matches 23 kinds and the repo only had templates for 13 of them. They are all plain minimal objects, and I dry ran each one against a 1.31 apiserver before writing any tests, so a failing test means the policy and not a bad manifest.

## C-0234, no external secret storage

This one is inverted compared to everything else in the library. It does not deny a workload for doing something bad, it denies it for not doing something good. So the expression is a bare `exists()` with no negation on it, and a workload with no volumes at all fails.

All three `has()` calls are needed and they have to be chained. `csi` is optional on a volume and `volumeAttributes` is optional on a csi source, so `has(volume.csi.volumeAttributes.secretProviderClass)` on a hostPath volume errors instead of returning false. There is a test for each of those shapes.

Worth saying plainly: this fires on nearly every workload in a normal cluster. The Rego does the same, CIS words it as "consider", so it is advisory rather than a hard rule. I put a section in the doc page saying to think about `validationActions: [Warn]` before `[Deny]`. Nothing to decide in the YAML since we do not ship bindings.

Parity note: kubescape derives no remediation path from this one, and I do not think it can. The failing term is a missing collection, there is no element to blame. The Rego emits a `YOUR_VALUE` placeholder fixPath, which we would not copy anyway.

## C-0295, duplicate env var names

Nine deny blocks in the Rego, one expression here off a `variables` block.

**ReplicationController is in this one and it is not in the others.** The Rego's workload set includes it, so leaving it out would drop coverage rather than add any. It is in the kind list in the variable and in matchConstraints. Flagging it because every other container walk in this repo stops at the five apps and batch kinds.

Finding a duplicate means comparing every entry against every other one in the same container, so this is quadratic however you write it, CEL has no set type to deduplicate with. `exists_one` is the cheapest way I found to say it: a name that appears once matches exactly one entry, itself, and a name that appears twice matches two and fails. `filter(...).size() == 1` says the same thing but builds a throwaway list per entry.

I measured it on 1.31 before opening this, since it is easily the most expensive expression in the repo:

- 400 env entries in one container is fine, 450 runs out of the runtime budget and the apiserver cancels with `operation cancelled: actual cost limit exceeded`
- the cost is per container and not per pod. 3 containers of 250 passes, 1 container of 450 does not, so the real ceiling is somewhere around 200k pairwise comparisons added up over the containers
- the policy itself is accepted at creation, so the compile time estimate is nowhere near the limit. Only the runtime budget bites
- with `failurePolicy: Fail` that means going over gets the request rejected with a cost error instead of the message

That is written up in the doc page as well. I did try hoisting `env.map(e, e.name)` out of the inner loop with `cel.bind()` to cut the constant factor, and the apiserver rejects the policy with `undeclared reference to 'names'`, so that is not available to us here.

Parity note: the derived remediation path is `spec.containers[N].env`. The container is pinned, the env index inside it is not, same as C-0207.

## C-0212, default namespace

17 near identical Rego rules, one expression. Every kind this matches is namespaced and reaches its namespace through the same path, so there is no kind dispatch guard. I think it is the only control in the repo where that is true.

Both Rego branches deny, and the one expression reproduces both. A missing `metadata.namespace` fails on purpose, because a manifest with no namespace goes to `default` when you apply it. At live admission the apiserver has already filled that field in before any policy runs, so that half only ever shows up when the object is read straight from a file, which is what a kubescape scan does.

This overlaps C-0061 on the seven workload kinds, with the identical expression. They are separate CIS recommendations and both exist upstream, so both exist here. C-0212 adds 16 more kinds on top, so a ConfigMap or a Role or an Ingress in `default` is caught here and not there. Bind both and a workload in `default` gets reported twice. There is a note about that in the doc page.

Getting 23 kinds and 10 apiGroups right is the fiddly part of this one, so the tests are one fail case per kind. A fail case is what proves matchConstraints actually matched that kind, a pass case would look the same whether the policy matched or not. Six pass cases on top for the non-default path.

Autoscaling is the one place I did not write `apiVersions: ["v1"]`. HorizontalPodAutoscaler is served under both v1 and v2 and clusters use v2, so matching only v1 there would quietly miss most of them.

Same operational warning as C-0234 in the doc page. `default` is where a lot of ordinary bootstrap happens, and binding this as Deny blocks more than you would expect, including updates to objects that were already living there before the policy went in.

Parity note: Rego gives the absent namespace case a fixPath carrying `YOUR_NAMESPACE`. We give a review path on `metadata.namespace` for both cases instead, since we do not invent placeholder values.

## Tests

61 tests, all green on kind v1.31. 15 for C-0234, 17 for C-0295, 29 for C-0212.

## Summary by CodeRabbit

- **New Features**
  - Added policies preventing duplicate environment-variable names in containers.
  - Added enforcement requiring supported workloads to use external secret storage.
  - Added protection against deploying supported resources to the `default` or unspecified namespace.

- **Documentation**
  - Added policy guidance, severity details, covered resources, and recommended enforcement modes.

- **Tests**
  - Added comprehensive validation coverage and Kubernetes test resources for all three policies.

### DIFF
--- a/controls/C-0234/policy.yaml
+++ b/controls/C-0234/policy.yaml
@@ -1,0 +1,54 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
+metadata:
+  name: "kubescape-c-0234-deny-workloads-without-external-secret-storage"
+  labels:
+    controlId: "C-0234"
+  annotations:
+    controlUrl: "https://kubescape.io/docs/controls/c-0234/"
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   [""]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["pods"]
+    - apiGroups:   ["apps"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["deployments","replicasets","daemonsets","statefulsets"]
+    - apiGroups:   ["batch"]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["jobs","cronjobs"]
+  variables:
+    # volumes is optional on every kind, and this control has to treat "no volumes at
+    # all" the same as "volumes, none of them a secrets-store CSI volume", so each
+    # branch defaults to an empty list rather than reading the field unguarded.
+    - expression: |
+        object.kind == 'Pod'
+          ? (has(object.spec.volumes) ? object.spec.volumes : [])
+          : object.kind in ['Deployment','ReplicaSet','DaemonSet','StatefulSet','Job']
+            ? (has(object.spec.template.spec.volumes) ? object.spec.template.spec.volumes : [])
+            : object.kind == 'CronJob'
+              ? (has(object.spec.jobTemplate.spec.template.spec.volumes) ? object.spec.jobTemplate.spec.template.spec.volumes : [])
+              : []
+      name: volumes
+  validations:
+    # This control is inverted compared to the rest of the library: it denies on the
+    # ABSENCE of something good rather than the presence of something bad. So the
+    # expression is a bare exists() with no negation, and an empty volumes list fails.
+    #
+    # All three has() calls are needed and they have to be chained. csi is optional on a
+    # volume and volumeAttributes is optional on a csi source, so
+    # has(volume.csi.volumeAttributes.secretProviderClass) on a hostPath volume errors
+    # instead of returning false.
+    - expression: >
+        variables.volumes.exists(volume,
+          has(volume.csi) &&
+          has(volume.csi.volumeAttributes) &&
+          has(volume.csi.volumeAttributes.secretProviderClass)
+        )
+      messageExpression: >
+        object.kind + '/' + object.metadata.name + ' does not read its secrets from an external store through the secrets-store CSI driver. (see more at https://kubescape.io/docs/controls/c-0234/)'
--- a/controls/C-0234/policy.yaml
+++ b/controls/C-0234/policy.yaml
@@ -40,15 +40,24 @@ spec:
     # ABSENCE of something good rather than the presence of something bad. So the
     # expression is a bare exists() with no negation, and an empty volumes list fails.
     #
-    # All three has() calls are needed and they have to be chained. csi is optional on a
-    # volume and volumeAttributes is optional on a csi source, so
+    # The has() calls are chained rather than collapsed. csi is optional on a volume and
+    # volumeAttributes is optional on a csi source, so
     # has(volume.csi.volumeAttributes.secretProviderClass) on a hostPath volume errors
-    # instead of returning false.
+    # instead of returning false. driver is required by the API but a manifest read off
+    # disk during a scan never went through that validation, so it is guarded too.
+    #
+    # The driver name and the non-empty check are both STRICTER than the Rego, which only
+    # asks whether the secretProviderClass key resolves. Without them any CSI driver that
+    # happens to expose an attribute of that name satisfies the control, and so does an
+    # empty value, neither of which is reading secrets from an external store.
     - expression: >
         variables.volumes.exists(volume,
           has(volume.csi) &&
+          has(volume.csi.driver) &&
+          volume.csi.driver == 'secrets-store.csi.k8s.io' &&
           has(volume.csi.volumeAttributes) &&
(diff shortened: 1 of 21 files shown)


controls/C-0234/policy.yaml
```diff
@@ -0,0 +1,54 @@
...
+    # All three has() calls are needed and they have to be chained. csi is optional on a
+    # volume and volumeAttributes is optional on a csi source, so
+    # has(volume.csi.volumeAttributes.secretProviderClass) on a hostPath volume errors
+    # instead of returning false.
+    - expression: >
+        variables.volumes.exists(volume,
+          has(volume.csi) &&
+          has(volume.csi.volumeAttributes) &&
+          has(volume.csi.volumeAttributes.secretProviderClass)
+        )
```
matthyx, line 61: Blocker: this check doesn't actually require the secrets-store CSI driver. It only checks that *some* CSI volume has a `volumeAttributes.secretProviderClass` key — it never compares `volume.csi.driver` against `secrets-store.csi.k8s.io`, and `has()` is satisfied even by an empty string. So a workload using an unrelated CSI driver that happens to expose an attribute literally named `secretProviderClass` (or sets it to `""`) passes the control, even though it isn't reading secrets from an external store at all. That contradicts both the doc page and the messageExpression, which both claim this is enforcing use of "the secrets-store CSI driver."

Suggest tightening to something like:
```
variables.volumes.exists(volume,
  has(volume.csi) &&
  volume.csi.driver == 'secrets-store.csi.k8s.io' &&
  has(volume.csi.volumeAttributes) &&
  has(volume.csi.volumeAttributes.secretProviderClass) &&
  volume.csi.volumeAttributes.secretProviderClass != ''
)
```
with a couple of new failing fixtures (unrelated CSI driver, empty `secretProviderClass`) added to `tests.json`.

controls/C-0295/policy.yaml
```diff
@@ -0,0 +1,60 @@
...
+  annotations:
+    controlUrl: "https://kubescape.io/docs/controls/c-0295/"
+spec:
+  failurePolicy: Fail
+  matchConstraints:
+    resourceRules:
+    - apiGroups:   [""]
+      apiVersions: ["v1"]
+      operations:  ["CREATE", "UPDATE"]
+      resources:   ["pods","replicationcontrollers"]
```
matthyx: Blocker: ephemeral containers never actually get checked in practice, even though the `containers` variable and the doc page both claim coverage of them. Kubernetes only allows adding ephemeral containers through the `pods/ephemeralcontainers` subresource (that's how `kubectl debug` works) — a resource rule that matches plain `pods` doesn't intercept requests to that subresource. So `object.spec.ephemeralContainers` is effectively always empty on the requests this policy actually sees.

Add `pods/ephemeralcontainers` to the resource list here:
```
resources:   ["pods","pods/ephemeralcontainers","replicationcontrollers"]
```
(No need to add a `tests.json` case for it — the harness drives `kubectl apply` and can't issue subresource requests.)

matthyx (changes requested):
Good set of three controls, and the fixture/test coverage for C-0212 and C-0295 is solid. Verified against the actual CI run (minikube 1.31.1) that the tests currently pass, and independently re-checked CodeRabbit's three flagged issues rather than taking them at face value:

- The C-0234 "bare `{name: test-volume}`" concern doesn't hold up — I pulled the CI log and that test genuinely passes against a real v1.31 apiserver (`kubectl --dry-run=server apply` returns 0), so no fix needed there.
- Two other gaps are real and worth fixing before merge (inline comments below): C-0234's `exists()` doesn't actually pin down the secrets-store CSI driver or reject an empty `secretProviderClass`, and C-0295's ephemeral-container handling is unreachable because the policy never matches the `pods/ephemeralcontainers` subresource.

Once those two are addressed (and tests added for the C-0234 case), this looks good to merge.

matthyx (approved):
Both blockers from the previous round are fixed, and I re-verified against the actual CI transcript rather than just the green check:

- C-0234 now pins `volume.csi.driver == 'secrets-store.csi.k8s.io'` and requires a non-empty `secretProviderClass`. The two new fail-case tests (unrelated driver, empty value) ran and passed against the real v1.31 minikube apiserver in this run.
- C-0295 now matches `pods/ephemeralcontainers`, so the ephemeral-container branch of the container walk is reachable in practice.

CI is green (test-all-policies, GitGuardian, DCO). LGTM, approving.

PR #118, opened 2026-08-13
add C-0225 and C-0296

Sanchit2662:

Two more. C-0225 is `ensure-default-service-accounts-has-only-default-roles` plus `automount-default-service-account`, C-0296 is `mismatching-selector`.

## C-0225

Two unrelated rules under one control ID, so two validations each with its own kind guard.

The bootstrap-label exemption is the part that matters. Kubernetes labels its own bindings `kubernetes.io/bootstrapping: rbac-defaults` and some of them bind the default ServiceAccount on purpose, so without it this fires on the cluster's own RBAC everywhere. There is a test for the exemption and one for a different bootstrapping value still being denied, so it stays an exact match.

An absent `automountServiceAccountToken` fails rather than passes. Kubernetes mounts when nothing says otherwise, same as C-0034.

15 tests, green on kind v1.31.

## C-0296

Three selector shapes so three validations: the usual `spec.selector`, CronJob's under `spec.jobTemplate.spec`, and ReplicationController's bare label map with no wrapper. All four `matchExpressions` operators are in.

**All the tests are passes, and I could not write a failing one.** The API server rejects a mismatching selector itself before any policy runs:

```
The Deployment "mismatch-deploy" is invalid: spec.template.metadata.labels:
  Invalid value: ...: `selector` does not match template `labels`
```

Same for all seven kinds. Job is rejected with and without `manualSelector`, CronJob will not take `manualSelector: true` at all. So there is no object the harness can hand this policy that it would be the one to reject.

The passing tests still earn their place, since with `failurePolicy: Fail` a CEL error here would deny every workload in the cluster. They cover all four operators and both of the selector shapes I have fixtures for. The ReplicationController branch has no fixture in this repo yet, so I covered it in the engine run below rather than adding one here and colliding with #117.

I tested the deny half through the kubescape CEL engine against manifest fixtures, 27 cases across all four operators and all three shapes, all correct.

So this one does nothing at admission and I would rather say that than hide it. It is useful when scanning manifests, because a file on disk never went through that validation. Tell me if you would rather not ship a policy that is redundant in a cluster and I will drop it.

12 tests, green on kind v1.31.

## Summary by CodeRabbit

* **New Features**
  * Added admission controls to prevent unauthorized use of the default ServiceAccount and unintended token automounting.
  * Added validation to detect mismatches between workload selectors and pod-template labels across supported Kubernetes workloads.
  * Registered both policies for deployment and included them in the policy library.

* **Documentation**
  * Added guidance, severity details, configuration information, and usage instructions for both controls.

### DIFF
--- a/controls/C-0296/tests.json
+++ b/controls/C-0296/tests.json
@@ -1,0 +1,81 @@
+[
+    {
+        "name": "Deployment whose selector matches its pod template labels is allowed",
+        "template": "deployment.yaml",
+        "expected": "pass"
+    },
+    {
+        "name": "Deployment with extra labels on the pod template is allowed, the selector only has to be a subset",
+        "template": "deployment.yaml",
+        "expected": "pass",
+        "field_change_list": [
+            "spec.template.metadata.labels={\"label\": \"test-deployment\", \"tier\": \"web\"}"
+        ]
+    },
+    {
+        "name": "Deployment selecting with an In matchExpression that its template satisfies is allowed",
+        "template": "deployment.yaml",
+        "expected": "pass",
+        "field_change_list": [
+            "spec.selector={\"matchExpressions\": [{\"key\": \"label\", \"operator\": \"In\", \"values\": [\"test-deployment\"]}]}"
+        ]
+    },
+    {
+        "name": "Deployment selecting with a NotIn matchExpression that its template satisfies is allowed",
+        "template": "deployment.yaml",
+        "expected": "pass",
+        "field_change_list": [
+            "spec.selector={\"matchLabels\": {\"label\": \"test-deployment\"}, \"matchExpressions\": [{\"key\": \"tier\", \"operator\": \"NotIn\", \"values\": [\"db\"]}]}"
+        ]
+    },
+    {
+        "name": "Deployment selecting with an Exists matchExpression that its template satisfies is allowed, and reading the absent values field must not error",
+        "template": "deployment.yaml",
+        "expected": "pass",
+        "field_change_list": [
+            "spec.selector={\"matchExpressions\": [{\"key\": \"label\", \"operator\": \"Exists\"}]}"
+        ]
+    },
+    {
+        "name": "Deployment selecting with a DoesNotExist matchExpression that its template satisfies is allowed",
+        "template": "deployment.yaml",
+        "expected": "pass",
+        "field_change_list": [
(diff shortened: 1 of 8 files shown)


controls/C-0296/tests.json
matthyx, line 1: Non-blocking: the third validation in `controls/C-0296/policy.yaml` (the `ReplicationController` branch, plain-map selector with no `matchLabels`/`matchExpressions` wrapper) has no test case here at all — `tests.json` only covers Deployment/ReplicaSet/DaemonSet/StatefulSet/Job/CronJob. `test-resources/replicationcontroller.yaml` already exists and already has a matching selector, so a trivial addition would close the gap and let `test-all-policies` actually exercise that branch on a real cluster (rather than relying only on the external kubescape-engine run described in the PR body):

```json
{
    "name": "ReplicationController whose selector matches is allowed",
    "template": "replicationcontroller.yaml",
    "expected": "pass"
}
```

matthyx (dismissed):
Reviewed both policies against the actual Kubernetes selector-matching semantics (`labels.Requirement.Matches` for `In`/`NotIn`/`Exists`/`DoesNotExist`) and against the existing kind-guard/`has()`-chain patterns used elsewhere in this repo (e.g. C-0034) — the CEL is correct and the guard ordering that avoids `NoSuchKey` on `e.values` for `Exists`/`DoesNotExist` is sound. Docs, README entry, and `kustomization.yaml` wiring all check out. CI (`test-all-policies`, GitGuardian, DCO) is green.

One non-blocking test-coverage gap noted inline: C-0296's `ReplicationController` validation branch isn't exercised by any test case even though the fixture already exists.

Approving — good to merge.

matthyx:
rebase pls

matthyx (approved):
Re-reviewed after the rebase — the diff for `controls/C-0225/**`, `controls/C-0296/**`, `README.md`, and `controls/kustomization.yaml` is byte-for-byte identical to what I approved earlier (only the surrounding rebase context in `kustomization.yaml` shifted), so my earlier findings stand: the CEL matches real Kubernetes selector/RBAC semantics, and the guard ordering is sound. My prior review got auto-dismissed by the push, so re-approving here.

`test-all-policies` is still running on the rebased commit `8898180` as of this review — worth confirming it comes back green before merging, though the code itself hasn't changed since it last passed. The non-blocking suggestion from before (add a `ReplicationController` test case to `controls/C-0296/tests.json`) still stands but isn't a blocker.

Approving — good to merge once CI on the rebased commit finishes green.

PR #119, opened 2026-08-14
add C-0202, C-0203 and C-0204

Sanchit2662:

Three more from the Pod Security Admission family, converted the same way as C-0198 to C-0201: each control becomes the workload check its CIS item is really about, not the namespace label the shared Rego rule looks at.

One correction up front. I went in assuming all three were new checks. Two are not, and I only caught it by reading the bundle instead of the control descriptions. **C-0203 is the same check as C-0048**, same field and same shape. **C-0204 overlaps C-0044.** I still wrote both, because the loader keys on the `controlId` label and a control with no policy directory of its own cannot report at all. C-0013 already overlaps C-0016 and C-0198 the same way. The doc pages spell out the real relationships so nobody has to rediscover this.

### C-0202, Windows hostProcess

The only genuinely new one. Checks `windowsOptions.hostProcess` on the pod security context and on every container, init and ephemeral included.

Two validations rather than one `&&`. Joined together, a container-level violation still reports the pod-level field as the thing to fix, which sends you to the wrong place.

What surprised me is that the API server gets there first. `hostProcess: true` requires `hostNetwork: true`, and if any container is a HostProcess container then all of them must be. So a mixed pod never reaches this policy on a live cluster, only in a scanned file. Every failing test carries `hostNetwork: true` for that reason.

### C-0203, hostPath volumes

Denies any `hostPath` volume outright. Written per kind against `object` rather than through a variable, matching C-0048, because the scan-side path derivation walks object-rooted expressions, so a failure can point at `spec.volumes[2].hostPath` instead of at nothing.

Worth separating from C-0045, which only denies a hostPath mounted *without* `readOnly: true`. There is a test for a read only hostPath mount: C-0045 passes it, this denies it.

### C-0204, hostPorts

Two deliberate differences from C-0044. It walks init and ephemeral containers, so a hostPort on an init container is denied here and passes C-0044. And `hostPort: 0` is allowed, since `HostPort` is a plain `int32` rather than a pointer and an explicit zero means the same as leaving it out. Tests for both.

### Testing

40 cases, all green. I checked every fixture against the API server before writing the test, which is how the hostProcess rules turned up. No new files in `test-resources/`, the existing templates cover all of it.

Small thing I hit: `scripts/change-yaml-field.py` cannot create a list that does not already exist, so where I needed a new `ports` or `volumes` list I passed the whole list as one JSON value. Happy to send a fix for the script separately.

## Summary by CodeRabbit

* **New Features**
  * Added admission policies that block Windows HostProcess containers, `hostPath` volumes, and nonzero container host ports.
  * Policies cover Pods and supported workload resources, including init and ephemeral containers where applicable.
* **Documentation**
  * Added policy guidance, severity details, supported resources, and implementation references for controls C-0202, C-0203, and C-0204.
* **Tests**
  * Added conformance coverage for allowed and rejected configurations across supported Kubernetes resource types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,9 @@ kubectl -n policy-example run nginx --image=nginx --restart=Never
 | [C-0212](https://kubescape.io/docs/controls/c-0212/) | The default namespace should not be used | [kubescape-c-0212-deny-resources-in-default-namespace](/docs/policies-based-on-kubescape-controls/kubescape-c-0212-deny-resources-in-default-namespace.md) | not configurable |
 | [C-0225](https://kubescape.io/docs/controls/c-0225/) | Prefer using dedicated EKS Service Accounts | [kubescape-c-0225-deny-default-service-account-rbac-and-automount](/docs/policies-based-on-kubescape-controls/kubescape-c-0225-deny-default-service-account-rbac-and-automount.md) | not configurable |
 | [C-0296](https://kubescape.io/docs/controls/c-0296/) | Mismatching selector | [kubescape-c-0296-deny-workloads-with-mismatching-selector](/docs/policies-based-on-kubescape-controls/kubescape-c-0296-deny-workloads-with-mismatching-selector.md) | not configurable |
+| [C-0202](https://kubescape.io/docs/controls/c-0202/) | Minimize the admission of Windows HostProcess Containers | [kubescape-c-0202-deny-windows-hostprocess-containers](/docs/policies-based-on-kubescape-controls/kubescape-c-0202-deny-windows-hostprocess-containers.md) | not configurable |
+| [C-0203](https://kubescape.io/docs/controls/c-0203/) | Minimize the admission of HostPath volumes | [kubescape-c-0203-deny-hostpath-volumes](/docs/policies-based-on-kubescape-controls/kubescape-c-0203-deny-hostpath-volumes.md) | not configurable |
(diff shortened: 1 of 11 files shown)


Sanchit2662:
Hey @matthyx , please give this one a review.

matthyx (approved):
Reviewed the CEL logic for all three policies (C-0202, C-0203, C-0204) and traced through every branch condition.

- **C-0202** (Windows hostProcess): the pod-level and container-level checks are correctly kept as two separate validations so the reported field points at the actual violation, not always the pod security context. The `podSpec`/`containers` variable ternaries cover every kind admitted by `matchConstraints` (Pod incl. ephemeralcontainers, Deployment/ReplicaSet/DaemonSet/StatefulSet/Job, CronJob) with no gap that would fall through to an empty map/list unexpectedly.
- **C-0203** (hostPath volumes): confirmed the expression is logically identical to the existing C-0048 check (verified against `controls/C-0048/policy.yaml`), and the short-circuit ordering (`object.kind != X || ...`) correctly avoids evaluating fields that don't exist on the wrong kind.
- **C-0204** (hostPorts): verified against `controls/C-0044/policy.yaml` that the described differences are accurate — C-0044 only checks `spec.containers` and fails on explicit `hostPort: 0` via bare `has()`, while this policy walks init/ephemeral containers too and correctly treats `hostPort: 0` as equivalent to absent (matches the Go `int32` zero-value semantics).

Docs follow the existing doc format exactly (spot-checked against C-0198's doc), README/kustomization entries are wired up, and CI (`test-all-policies`, 40 new cases) is green. No blockers. Approving.

PR #120, opened 2026-08-16
fix C-0020 denying everything with a hostPath when cloudProvider is not configured

### LINKED ISSUE #116: C-0020 denies everything with a hostPath because cloudProvider is missing from the shipped config
I hit this while testing the kubescape CEL engine against a real cluster, and I think it is a problem in the library itself and not on our side.

`configuration/basic-control-configuration.yaml` has no `cloudProvider` key. C-0020 is the only policy that reads `params.settings.cloudProvider`, and I checked, it is still missing on main today.

So if you deploy the config this repo ships and bind C-0020, the expression hits a key that is not there and errors at runtime. Every policy here is `failurePolicy: Fail`, all 54 of them, so the apiserver does not skip it, it denies the request.

## Why it is easy to miss

The expression is a chain of `params.settings.cloudProvider != 'eks' || ...` clauses, and CEL short circuits. It only gets as far as reading the key once a hostPath volume actually matches one of the sensitive paths. Anything else passes cleanly and never touches it.

Which means the control looks completely healthy until the exact moment it is supposed to catch something, and then it fails for the wrong reason.

I ran all 8 of the C-0020 fixtures through it and the split is exact. The 4 that should pass evaluate fine with no errors. The 4 that should fail all error.

CI does not see it because `test-resources/default-control-configuration.yaml` sets `cloudProvider: aks`, so the harness always runs with the key present. The shipped config is the one that is missing it.

## Showing it is really the missing key

Same pod, same policy, I only changed the params:

| cloudProvider | Pod with hostPath `/etc/kubernetes` | clean Pod |
|---|---|---|
| missing (what is shipped) | denied | admitted |
| `aks` | denied, this is the real violation | admitted |
| `gke` | admitted | admitted |
| `""` | admitted | admitted |

The missing row denying is the failurePolicy path, not the control doing its job. Tested on kind v1.31.0 with a Deny binding.

## What I think the fix is

Add `cloudProvider: ""` to `configuration/basic-control-configuration.yaml`.

I tested that one and it is safe. With an empty string every clause short circuits to true, so nothing gets denied, and anyone actually running on AKS or EKS sets it and the control works properly. No false positives either way.

The other thing worth considering, though this is more your call: guard the read in the policy itself, something like `!has(params.settings.cloudProvider) || params.settings.cloudProvider != 'aks' || ...`. Adding the key to the shipped config fixes the default path, but a user who writes their own ControlConfiguration and leaves `cloudProvider` out would still walk into the same deny. The guard covers that, the config change does not.

Happy to send a PR for either or both, just say which you prefer.

Small related note, the same missing key means kubescape's scan cannot report a C-0020 violation at all right now, it just errors and marks the resource skipped. I opened kubescape/kubescape#3059 about the skip-versus-deny side of that, which is a separate thing on our end.

Sanchit2662:
Fixes #116.

`cloudProvider` is missing from `configuration/basic-control-configuration.yaml`, and C-0020 is the only policy that reads it. Reading a key that is not there is an eval error, and every policy here is `failurePolicy: Fail`, so the apiserver denies the request instead of skipping the policy. CEL short circuits, so the key only gets read once a hostPath volume actually matches one of the sensitive paths. The control looks completely healthy until the moment it is supposed to catch something.

I did both fixes from the issue, because they cover different holes.

Added `cloudProvider: ""` to the shipped config. With an empty value every clause short circuits to true, so nothing gets denied, and anyone actually on AKS or EKS sets it and the control works properly.

Added a `cloudProvider` variable on the policy that falls back to `''` when the key is absent, and pointed the 9 reads at it. The config change fixes the default path, but a user who writes their own ControlConfiguration and leaves the key out would still walk into the same deny. This covers that one.

On the test side, there was no way to run a control against a params file missing a key, which is why CI never saw this. `run-control-tests.py` now takes an optional `param_template` per test, same shape as the `binding_template` option that was already there, defaulting to the current file so no other control changes. Added `test-resources/empty-control-configuration.yaml` and two C-0020 cases that point at it.

Checked on kind v1.31. All 10 C-0020 cases pass, and the two new ones both fail against the policy as it is today, so they do catch the bug rather than just sitting there.

## Summary by CodeRabbit

* **New Features**
  * Added optional cloud provider configuration for EKS, AKS, and GKE environments.
  * Cloud-specific credential path checks are skipped when no provider is configured.

* **Bug Fixes**
  * Prevented validation errors when cloud provider settings are missing.
  * Improved handling of Pod, workload, and CronJob credential-path checks.

* **Tests**
  * Added coverage for missing cloud provider configuration and Kubernetes host paths.
  * Test cases can now use custom parameter templates.

### DIFF
--- a/configuration/basic-control-configuration.yaml
+++ b/configuration/basic-control-configuration.yaml
@@ -3,6 +3,9 @@ kind: ControlConfiguration
 metadata:
   name: basic-control-configuration
 settings:
+  # Set this to eks, aks or gke to make C-0020 check that cloud's credential
+  # paths. An empty value means "unknown cloud", so C-0020 checks none of them.
+  cloudProvider: ""
   cpuLimitMax: 5
   cpuLimitMin: 0.5
   cpuRequestMax: 5
--- a/controls/C-0020/policy.yaml
+++ b/controls/C-0020/policy.yaml
@@ -25,12 +25,20 @@ spec:
       apiVersions: ["v1"]
       operations:  ["CREATE", "UPDATE"]
       resources:   ["jobs","cronjobs"]
+  variables:
+    # cloudProvider is optional in the ControlConfiguration, and reading a key
+    # that is not there is an eval error, which failurePolicy: Fail turns into a
+    # denial. Falling back to an empty string keeps that a pass instead: no
+    # provider matches, so none of the credential path lists are checked.
+    - name: cloudProvider
+      expression: >
+        has(params.settings.cloudProvider) ? params.settings.cloudProvider : ''
   validations:
     - expression: >
         object.kind != 'Pod' || !has(object.spec.volumes) || object.spec.volumes.all(vol,
         !has(vol.hostPath) || !has(vol.hostPath.path) ||
         (
-          (params.settings.cloudProvider != 'eks' ||
+          (variables.cloudProvider != 'eks' ||
           ["/.aws/","/.aws/config/","/.aws/credentials/"].all(unsafePath,
           (
             (
@@ -42,7 +50,7 @@ spec:
               unsafePath != vol.hostPath.path
             )))) &&
 
-          (params.settings.cloudProvider != 'aks' ||
+          (variables.cloudProvider != 'aks' ||
           ["/etc/","/etc/kubernetes/","/etc/kubernetes/azure.json","/.azure/","/.azure/credentials/","/etc/kubernetes/azure.json"].all(unsafePath,
           (
             (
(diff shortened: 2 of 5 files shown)


matthyx (approved):
Verified this fixes #116: the `settings.cloudProvider` field is already declared (`type: string`) in `configuration/policy-configuration-definition.yaml`, so both the shipped-config default and the CEL-side `has()` fallback are consistent with the CRD schema and won't diverge from it.

Both parts of the fix are correct and complementary:
- `cloudProvider: ""` in `basic-control-configuration.yaml` fixes the shipped default.
- The `variables.cloudProvider` fallback (`has(...) ? ... : ''`) in `controls/C-0020/policy.yaml` covers user-authored `ControlConfiguration`s that omit the key entirely — matches the existing style of unguarded `params.settings.X` reads elsewhere, since `settings` itself is always present.

Test coverage is solid: the new `empty-control-configuration.yaml` fixture plus two new pass-case tests actually exercise the missing-key path (existing `default-control-configuration.yaml` already hardcodes `cloudProvider: aks`, which is exactly why CI never caught this). `run-control-tests.py`'s new `param_template` option is backward compatible (defaults to the previous hardcoded file). `test-all-policies` CI is green.

No blockers. Approving.

PR #121, opened 2026-08-17
add C-0193, C-0194, C-0195 and C-0197

Sanchit2662:

This is the last of the Pod Security Admission controls, and three of the four are checks the library already has. That is on purpose. A policy carries one `controlId` label, so a control with no policy of its own is not reported at all, no matter how many other policies happen to look at the same field. C-0013 already overlaps C-0016 and C-0198 for the same reason.

**C-0193, privileged containers.** The one that is not a copy. C-0057 checks the same `privileged` field but also denies a container adding `SYS_ADMIN` or `ALL`, so it denies strictly more. I kept this one to the privileged flag only, which is what its CIS item is about. There is a test for exactly that difference: same fixture and same field change as C-0057's SYS_ADMIN case, expecting pass here instead of fail.

**C-0197, allowPrivilegeEscalation.** Same check as C-0016, same reading, including that leaving the field out is a denial and not a pass. Kubernetes defaults it to true, so unset is the risk the control is about.

**C-0194 and C-0195, hostPID and hostIPC.** Copies of C-0275 and C-0276. I copied the expressions rather than rewriting them, and left a comment on each saying to keep the pair identical if either one changes.

All four use the container list or podSpec variable shape that is already standard in the bundle, so init and ephemeral containers are covered where that makes sense.

Tests are 55 cases across the four, run against a real apiserver on kind v1.31, all passing. The hostPID and hostIPC suites are the C-0275 and C-0276 suites unchanged, so the two pairs are tested the same way.

Each docs page says which existing policy overlaps and how, rather than leaving someone to work that out from a scan report with two findings on one field.

## Summary by CodeRabbit

* **New Features**
  * Added admission policies blocking privileged containers and unsafe privilege escalation settings.
  * Added protections against workloads sharing host PID or IPC namespaces.
  * Policies cover Pods and common workload resources, including init and ephemeral containers.

* **Documentation**
  * Added policy documentation, configuration details, severity information, and implementation guidance.
  * Updated the policy library with four new Kubescape controls.

* **Tests**
  * Added comprehensive coverage for allowed and denied configurations across supported Kubernetes resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,10 @@ kubectl -n policy-example run nginx --image=nginx --restart=Never
 | [C-0202](https://kubescape.io/docs/controls/c-0202/) | Minimize the admission of Windows HostProcess Containers | [kubescape-c-0202-deny-windows-hostprocess-containers](/docs/policies-based-on-kubescape-controls/kubescape-c-0202-deny-windows-hostprocess-containers.md) | not configurable |
 | [C-0203](https://kubescape.io/docs/controls/c-0203/) | Minimize the admission of HostPath volumes | [kubescape-c-0203-deny-hostpath-volumes](/docs/policies-based-on-kubescape-controls/kubescape-c-0203-deny-hostpath-volumes.md) | not configurable |
 | [C-0204](https://kubescape.io/docs/controls/c-0204/) | Minimize the admission of containers which use HostPorts | [kubescape-c-0204-deny-containers-using-host-ports](/docs/policies-based-on-kubescape-controls/kubescape-c-0204-deny-containers-using-host-ports.md) | not configurable |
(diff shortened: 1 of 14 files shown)


matthyx (approved):
Verified the four new controls against the ones they build on:

- **C-0193**: correctly narrower than C-0057 — checks only `securityContext.privileged`, while C-0057 also denies `SYS_ADMIN`/`ALL` capability additions. Test 5 (pod-for-list-items.yaml, SYS_ADMIN without `privileged`) confirms the intended pass/fail split.
- **C-0194**/**C-0195**: expressions are byte-identical to C-0275/C-0276 respectively — confirmed by diffing against the current `main` files.
- **C-0197**: same container-list variable and same `allowPrivilegeEscalation == false` requirement as C-0016 (absent field = deny), matching description.

Other checks:
- CI (`test-all-policies`, GitGuardian, DCO) all green.
- `pod-for-list-items.yaml` template used in C-0193's tests already exists in the repo (used by C-0057's suite).
- README/kustomization additions are consistent with existing conventions (table isn't strictly ordered, appending is fine).
- No open CodeRabbit findings.

No blockers. LGTM.

PR #122, opened 2026-08-18
fix doc links, missing doc pages and list creation in change-yaml-field.py

Sanchit2662:
I went to fix two broken doc links and found a few more things while I was in there.

### Docs

- **C-0038** the file on disk is spelled `previleges`. Renaming it fixes the broken link and makes the filename match `metadata.name`, so two of the things on my list were really one bug.
- **C-0020** had no doc page at all. The README already linked to the right name, the file just was not there. I wrote it, including a note that the shipped `cloudProvider` is an empty string so the policy allows everything until you configure it.
- **C-0016** filename did not match `metadata.name`. Renamed, README updated.
- **C-0013** was the bad one. The file is a byte for byte copy of the C-0017 page, so C-0013 had no real docs at all. I wrote it from the policy and dropped the copy.
- **C-0004** has a policy and a doc page but no row in the library table, so it was invisible there. Added.
- One `hhttps://` typo in the usage example.

After this: 60 controls, 60 doc pages, 60 README links, nothing broken and every filename matches its `metadata.name`.

### scripts/change-yaml-field.py

The script cannot create a list that does not already exist, so `spec.initContainers.[0].ports.[0].hostPort=9000` dies with an IndexError when the container has no `ports` yet. I hit this writing the C-0204 tests and worked around it by passing the whole list as JSON.

Three things going on. The lookahead does `int('[0]')`, which always raises, so it creates a dict where a list is needed. `fields.index(f)` looks up the wrong element anyway, and by then `f` is an int so it is searching a list of strings for a number. And `root = []` builds a list that is not attached to the document, which is what actually throws.

I rewrote `apply_field` to pick the container type from the next field and always write through the parent.

To check I had not broken anything I ran the old and new versions over every `field_change_list` in the repo, 463 of them, and diffed the results. 462 identical, 0 regressions, 1 that only works now. That one is the C-0204 init container test, which I converted back to the normal form. Ran it on a cluster, 13/13 green.

That comparison also caught a regression I had introduced: I tightened the field name check to a full match and it broke `top-secret-allowed` in the C-0012 tests. The original was a prefix match so hyphenated keys always passed. Put it back with a comment.

## Summary by CodeRabbit

- **Documentation**
  - Corrected policy links and added documentation for controls C-0004, C-0013, C-0016, C-0020, and C-0038.
  - Clarified policy scope, severity, configuration requirements, and supported Kubernetes resources.
  - Removed outdated C-0013 documentation.

- **Bug Fixes**
  - Improved YAML field updates with better path validation, list expansion, Kubernetes key handling, and clearer errors.
  - Fixed port field updates so container and host ports are changed independently.


matthyx (approved):
Reviewed the diff against the actual CEL policies and metadata.

- `change-yaml-field.py`: traced `apply_field` by hand for the new `spec.initContainers.[0].ports.[0].containerPort=9000` / `...hostPort=9000` case in `controls/C-0204/tests.json` — it correctly grows the list, creates the intermediate dict, and writes through the parent reference rather than a detached local. The `top-secret-allowed` prefix-match fix is preserved (`re.match`, not `re.fullmatch`), so C-0012's tests keep passing. `test-all-policies` is green, consistent with the 463/463 comparison and 13/13 cluster run described in the PR body.
- Doc content checked against the live `policy.yaml` CEL expressions: C-0013's "both allowPrivilegeEscalation=false AND (runAsNonRoot or runAsUser≠0), container overrides pod" description matches the validation expression exactly; C-0020's per-cloud-provider hostPath list and the "empty cloudProvider allows everything" note match; C-0004's memory request/limit range check matches.
- All renamed/new doc filenames now match their `metadata.name` (C-0013, C-0016, C-0020, C-0038), and the C-0016 README link/hhttps typo fixes are correct.

Only non-blocking note: CodeRabbit's two flagged doc-wording issues (C-0016 title still says "not set" instead of "must be false"; C-0038 wording) are in files that were only renamed here, content untouched — pre-existing, not introduced by this PR. Fine to leave for a follow-up.

No blockers found — approving.

PR #123, opened 2026-08-20
fix: correct SYS_ADM typo in shared test fixtures

### LINKED ISSUE #108: Cleanup: Remove invalid SYS_ADM capability strings from shared test fixtures
During the fix for PR #104, we corrected the `SYS_ADM` typo in the C-0057 policy. However, the invalid string still lingers in several shared YAML test fixtures under `test-resources/`.

Specific files include:
* `pod-capabilities.yaml`
* `deployment-with-common-label-1.yaml`
* `deployment-with-common-label-2.yaml`

While this is technically harmless for the controls currently using them (since they assert on "any capability added" rather than the specific name), it should be cleaned up so the typo isn't copied forward into future tests.

ArneshBanerjee:
Fixes #108.

PR #104 corrected the SYS_ADM typo in the C-0057 policy, but the invalid string was left behind in the shared test fixtures. This renames it to SYS_ADMIN in the three files listed in the issue.

I left test-resources/pod-for-list-items.yaml and test-resources/deployment-for-list-items.yaml alone on purpose, even though they contain the same string. pod-for-list-items.yaml is used by the C-0046 test "Pod having containers with capabilities set but not from insecure list", which expects a pass. SYS_ADMIN is in the insecureCapabilities list in configuration/basic-control-configuration.yaml, so renaming it there would make that test fail. The typo is load bearing in that file.

For the three files changed here, nothing that reads them matches on capability names. C-0077 only looks at labels, C-0200 only checks whether any capability was added at all, and the C-0046 case on pod-capabilities.yaml already fails on NET_RAW so it stays failing either way.

I could not run the control test suite locally since it needs a live cluster, so I am relying on CI for that.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected container and pod capability configuration to use the valid `SYS_ADMIN` capability.
  * Updated deployment test configurations for consistent capability handling.

### DIFF
--- a/test-resources/deployment-with-common-label-1.yaml
+++ b/test-resources/deployment-with-common-label-1.yaml
@@ -23,4 +23,4 @@ spec:
         securityContext:
           capabilities:
             add:
-            - SYS_ADM
+            - SYS_ADMIN
--- a/test-resources/deployment-with-common-label-2.yaml
+++ b/test-resources/deployment-with-common-label-2.yaml
@@ -23,4 +23,4 @@ spec:
         securityContext:
           capabilities:
             add:
-            - SYS_ADM
+            - SYS_ADMIN
--- a/test-resources/pod-capabilities.yaml
+++ b/test-resources/pod-capabilities.yaml
@@ -13,5 +13,5 @@ spec:
     securityContext:
       capabilities:
         add:
-        - SYS_ADM
+        - SYS_ADMIN
         - NET_RAW


matthyx (approved):
Verified the reasoning in the PR description against the actual test suite rather than taking it on faith:

- `deployment-with-common-label-1.yaml` / `-2.yaml` are only referenced by `controls/C-0077/tests.json`, whose tests check labels, never capability names — unaffected by the rename.
- `pod-capabilities.yaml` is referenced by `C-0046` ("NET_RAW capability is blocked") and `C-0200` ("added capabilities is denied") — both keyed on `NET_RAW`/"any capability added", not the `SYS_ADM`/`SYS_ADMIN` string.
- For the two files intentionally left alone (`pod-for-list-items.yaml`, `deployment-for-list-items.yaml`), every test that cares about the exact capability name (`C-0057`, `C-0193`) overrides it explicitly via `field_change_list`, so the base fixture value doesn't matter there. The one test that *does* depend on the raw fixture value — `C-0046`'s "capabilities set but not from insecure capabilities list is allowed" — would indeed flip to a false failure if renamed, since `SYS_ADMIN` is in `insecureCapabilities` in `configuration/basic-control-configuration.yaml`. Leaving the typo in these two files is correct and load-bearing, exactly as described.

No blockers. `test-all-policies` CI has since gone green along with the rest of the checks. Approving.

PR #124, opened 2026-09-09
fix: scope C-0231 to LoadBalancers with a matchCondition

Sanchit2662:
Scopes C-0231 to actual LoadBalancers with a matchCondition, so every other Service is excluded from the control instead of quietly passing it and inflating the score.

## Summary by CodeRabbit

* **Changes**
  * Validation now applies only to Services configured with the `LoadBalancer` type.
  * Services without an explicit type or with another type are excluded from this validation scope.

### DIFF
--- a/controls/C-0231/policy.yaml
+++ b/controls/C-0231/policy.yaml
@@ -14,6 +14,20 @@ spec:
       apiVersions: ["v1"]
       operations:  ["CREATE", "UPDATE"]
       resources:   ["services"]
+  matchConditions:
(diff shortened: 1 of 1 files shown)


matthyx (approved):
Reviewed the LoadBalancer-only match condition against the existing C-0231 Rego filter and the offline CEL match-condition semantics. The guarded expression cleanly excludes missing/non-LoadBalancer Service types while retaining validation for LoadBalancers. Release overlays render successfully, and the PR checks are green. No blockers found.

Issue #126, opened 2026-09-12
C-0081: what should the matchCondition narrow on, the Rego filter or the policy's own argocd check?


Sanchit2662:
C-0231 and C-0225 both got a `matchCondition` so that resources the control was never about stop entering it and passing vacuously. C-0081 needs the same thing, but unlike those two there is no obvious expression to use, because the Rego rule and this policy do not agree on what counts as an Argo CD image.

Right now `matchConstraints` says `deployments`, so every Deployment in a cluster enters this control. Almost all of them pass, since the validation is vacuously true for anything that is not an Argo CD image, but they are still counted as having satisfied a CVE control that has nothing to do with them.

The two candidates disagree in two places.

The regolibrary filter (`rules/CVE-2022-24348/filter.rego`) takes any Deployment where some container image contains `argocd:v`. It only looks at `spec.template.spec.containers`.

This policy is stricter. It requires `/argocd:v` or an image starting with `argocd:v`, deliberately, so that an unrelated image like `myargocd:v1.8.7` is not treated as Argo CD. It also walks `initContainers` as well as `containers`.

So:

- Mirror the filter, and a Deployment running `myargocd:v1.8.7` enters the control and passes, because the validation excuses it. That is the case the policy already went out of its way to exclude, so we would be putting it back into the denominator.
- Mirror the policy, and the matchCondition and the validation use the same predicate, so nothing enters scope that the validation is only going to excuse. The cost is that the CEL denominator is then slightly smaller than the Rego one, on exactly the set of images Rego matches by accident.

I lean towards mirroring the policy. Having the gate and the check disagree is the thing that creates these vacuous passes in the first place, and the filter's looseness looks like an oversight rather than a decision. But it is a real divergence from Rego either way and I would rather it be written down than picked quietly.

Whichever we go with, the same question covers `initContainers`: the filter ignores them, the policy does not. An Argo CD image in an initContainer is a real thing to catch, so I would keep the policy's behaviour there too.

PR #127, opened 2026-09-12
fix: scope the C-0225 ServiceAccount rule to the default account with a matchCondition

Sanchit2662:
Scopes the ServiceAccount half of C-0225 to the default account with a matchCondition, so every other ServiceAccount stops passing the control vacuously; the kind guard keeps RoleBindings and ClusterRoleBindings untouched, since rule 1 is not name-narrowed.

## Summary by CodeRabbit

- **Policy Updates**
  - ServiceAccount validation now applies only to the `default` ServiceAccount.
  - RoleBinding and ClusterRoleBinding validation remains unchanged.
  - ServiceAccounts without a defined name are handled safely.


matthyx (approved):
Reviewed `79a89692029c325d79c08001ee7120b158f88353` against `main` at `1fbebccac3add584448ff5530db73414853a1d3c`.

The change is needed and sound. On the base policy, a non-default ServiceAccount enters C-0225 and both validations return true, creating a vacuous pass. The new guarded match condition returns false for that object while still returning true for the `default` ServiceAccount, RoleBindings, and ClusterRoleBindings. Kubescape's downstream evaluator applies `matchConditions` before validations and records a false gate as `Applicable: false`, so the change fixes the scoring scope rather than only changing admission behavior.

Related history: [#118](https://github.com/kubescape/cel-admission-library/pull/118) introduced and merged C-0225; its maintainer approval accepted the two-kind policy structure. Bot concerns there about bootstrap-label bypass, ServiceAccount groups, and controller-created defaults were not adopted before merge and concern pre-existing policy behavior, not this scoping diff. [#124](https://github.com/kubescape/cel-admission-library/pull/124) is the directly relevant merged precedent for using a guarded `matchCondition` to eliminate vacuous passes. [#126](https://github.com/kubescape/cel-admission-library/issues/126) discusses the same denominator problem for C-0081 but has different predicate ambiguity and does not supersede this PR. Searches covered GitHub-indexed open, closed, and merged PRs/issues for `C-0225`, `default ServiceAccount`, `matchConditions`, `vacuously`, and affected resource/name terms; no duplicate fix was found, but search indexing is not proof of absence.

Findings: no blockers. One optional LOW coverage improvement: `controls/C-0225/tests.json` observes admission allow/deny, so its non-default-ServiceAccount case cannot distinguish exclusion from a vacuous pass. A direct match-result regression assertion would protect this exact behavior.

Validation: `git diff --check` passed; YAML parsing and all six repository kustomization renders passed; eight non-cluster Python tests passed; six direct CEL predicate cases passed (default/dedicated/nameless/generated-name ServiceAccounts and both binding kinds). The base validations reproduced the vacuous pass for a dedicated ServiceAccount, while the new gate returned false. GitHub's Kubernetes v1.31.1 `test-all-policies` job and all other reported checks passed for this head. I did not rerun the cluster-mutating harness locally; its result is covered by the reviewed-head CI job. Independent implementation review recommends APPROVE and architecture review is CLEAR.

Verdict: APPROVE. No material correctness, security, compatibility, performance, or lifecycle regression was found. The PR was rechecked immediately before this review: open, non-draft, mergeable, unchanged head/base, no unresolved review threads, and green checks.

PR #128, opened 2026-09-17
fix: target Agent Sandbox v1beta1 APIs

dakshhhhh16:
## What this changes

Agent Sandbox v1.0 removed the legacy `v1alpha1` APIs from both `agents.x-k8s.io` and `extensions.agents.x-k8s.io`. A few of the Agent Runtime admission policies still listed both API versions in their match rules.

This updates C-0311, C-0312, C-0313, and C-0314 to match the supported `v1beta1` Agent Sandbox resources only. It also removes `v1alpha1` from the SandboxTemplate test CRD so the test setup reflects a current Agent Sandbox installation.

The `WorkerPool` rules continue to use `ate.dev/v1alpha1` because that is a separate Agent Substrate API and has not moved with the Agent Sandbox CRDs.

The documentation now records this API split and explains why `SandboxClaim` and `SandboxWarmPool` are not matched directly. The posture fields checked by these controls live on the referenced `SandboxTemplate` and the resulting `Sandbox`.

## Why

Leaving removed API versions in the policy definitions makes the admission bundle look compatible with resources that Agent Sandbox no longer serves. Keeping the rules aligned with the v1.0 API surface makes their supported scope clear and keeps the test fixtures representative of real installations.

Reference: https://github.com/kubernetes-sigs/agent-sandbox/releases/tag/v1.0.0

## Testing

- `python3 -m unittest discover -s scripts -p 'test_agent_test_crds.py'`
- Parsed the policy and test CRD YAML files with PyYAML
- `git diff --check`

## Summary by CodeRabbit

- **Policy Updates**
  - Admission policies for Sandboxes and SandboxTemplates now apply only to the `v1beta1` API.
  - Existing validation for image digests, networking, and resource operations remains unchanged.

- **Documentation**
  - Clarified API-version requirements for Agent Sandbox resources.
  - Documented why SandboxClaim and SandboxWarmPool resources are not matched directly.

- **Test Resources**
  - Updated the SandboxTemplate test definition to serve and store only `v1beta1`.


matthyx (approved):
Reviewed at `0089441` against `main` (`1ac05d2`).

**Need:** Confirmed. Agent Sandbox v1.0.0 removed `v1alpha1` from both `agents.x-k8s.io` and `extensions.agents.x-k8s.io` (release notes, kubernetes-sigs/agent-sandbox#1470), and `C-0297`/`C-0309` already matched `v1beta1` only, so this makes the bundle internally consistent. `WorkerPool` correctly keeps `ate.dev/v1alpha1`, which is unaffected.

**Correctness:** Match rules, test CRD, and doc changes are consistent. The `test_agent_test_crds.py` expectation of three fixture CRDs still holds. The new doc note about `SandboxClaim`/`SandboxWarmPool` is accurate — in v1.0 `SandboxClaimSpec` carries only `warmPoolRef`, lifecycle, pod metadata, env and volume claims; images, limits, runtime class and networking live on `SandboxTemplate`/`Sandbox`, and controller-created `Sandbox` objects still go through admission.

**Validation:**
- `python3 -m unittest discover -s scripts -p 'test_agent_test_crds.py'` — 8 tests OK
- C-0311 (12), C-0312 (19), C-0313 (66), C-0314 (4) cases via `run-control-tests.py` on a disposable kind cluster (Kubernetes 1.37) — all passed
- `kubectl kustomize apis/k8s-v1/`, PyYAML parse, `git diff --check` — OK
- CI `test-all-policies` is green on this commit

**Non-blocking note:** clusters still on Agent Sandbox v0.5.x whose clients write via `v1alpha1` lose admission coverage for these four controls. That matches what `C-0297`/`C-0309` already do, so it's fine here; the regolibrary rule metadata for the corresponding rules still lists `v1alpha1`, which could be aligned in a follow-up for consistency.

No blockers. Approving.

